PR #14, opened 2018-10-12
Set spi chunk limit on linux with cfg!

### LINKED ISSUE #13: Error: Os { code: 90, kind: Other, message: "Message too long" }', libcore/result.rs:945:5
Since cd980332041d19a9a0094885842743d6178c6975 there seems to happen the following error when trying to transfer bigger blocks of data:
`Os { code: 90, kind: Other, message: "Message too long" }', libcore/result.rs:945:5` 

This should be connected to the change from transfering single bytes to bigger byte buffers over spi.
(https://raspberrypi.stackexchange.com/questions/65595/spi-transfer-fails-with-buffer-size-greater-than-4096)

Caemor:
Spi.write now uses a max byte chunk size of 4096 on linux device because of a default defined max byte transfer size of 4096 on most linux devices (https://raspberrypi.stackexchange.com/questions/65595/spi-transfer-fails-with-buffer-size-greater-than-4096)

Fixes #13

### DIFF
--- a/src/interface.rs
+++ b/src/interface.rs
@@ -103,6 +103,7 @@ where
 
         // transfer spi data
         // Be careful!! Linux has a default limit of 4096 bytes per spi transfer
+        // see https://raspberrypi.stackexchange.com/questions/65595/spi-transfer-fails-with-buffer-size-greater-than-4096
         if cfg!(target_os = "linux") {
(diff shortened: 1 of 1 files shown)


Caemor:
Tested and it's now working once more

PR #17, opened 2018-10-14
Edition 2018

Caemor:
Updates library to 2018-edition.

When this update lands version 0.2 will be committed to crates.io.


Caemor:
Too many changes happened in between. Will make a fresh PR for the 2018-edition after 0.2 release

PR #22, opened 2018-11-20
Fix repository name

therealprof:


caemor:
Thanks :-)

caemor:
Also if you test this library tell me if anything is not working as expected/ more complicated/ counterintuitive ... :-)

therealprof:
@caemor I had troubles getting the examples to work. Other things I noted is that the construction of the display is rather cumbersome because it requires a lot of elements to be passed in which I don't really want to use (like CS and RESET).

Other than that I was pleasantly surprised how painless the whole thing was to setup (it literally worked on the first attempt without any troubles from the hardware or the rather little tested HAL implementation).

Also the binary size is much smaller compared to a SSD1306 hello world example, despite using graphics and whatnot. I'm very impressed.

PR #27, opened 2019-04-04
Make a few small updates to support TeXitoi rusty clock

caemor:


caemor:
Still merging since it only failed because clippy is unavailable on nightly atm

PR #31, opened 2019-08-11
Update for embedded-graphics 0.5.2

dbr:
Small update for 0.5.x which changed the draw trait method slightly,

https://github.com/jamwaffles/embedded-graphics/blob/master/CHANGELOG.md#050

### DIFF
--- a/examples/epd1in54_no_graphics/Cargo.toml
+++ b/examples/epd1in54_no_graphics/Cargo.toml
@@ -8,5 +8,5 @@ edition = "2018"
 
 epd-waveshare = { path = "../../", default-features = false, features = ["epd1in54"]}
 
-linux-embedded-hal = "0.2.2"
+linux-embedded-hal = "0.5.2"
 embedded-hal = { version = "0.2.2", features = ["unproven"] }
--- a/examples/epd1in54_no_graphics/Cargo.toml
+++ b/examples/epd1in54_no_graphics/Cargo.toml
@@ -8,5 +8,5 @@ edition = "2018"
 
 epd-waveshare = { path = "../../", default-features = false, features = ["epd1in54"]}
 
-linux-embedded-hal = "0.5.2"
+linux-embedded-hal = "0.2.2"
 embedded-hal = { version = "0.2.2", features = ["unproven"] }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -30,7 +30,7 @@ type_a_alternative_faster_lut = []
 
 [dependencies.embedded-graphics]
 optional = true
-version = "0.4.3"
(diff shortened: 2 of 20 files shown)


dbr:
Looks like the errors are all due to the InputPin and OutputPin methods being deprecated in [favour of failable `Result<...>` methods](https://github.com/rust-embedded/embedded-hal/blob/master/CHANGELOG.md#v023---2019-05-09)

I'll look and see how hard this is to address when I have time!

caemor:
First, thanks for you PR.

The error doesn't come from the warnings but you forgot to update the other two display files as well. Just adding what you did for the epd4in2 to them as well should solve the problem.
```
error[E0276]: impl has stricter requirements than trait

  --> src/epd1in54/graphics.rs:26:5

   |

26 | /     fn draw<T>(&mut self, item_pixels: T)

27 | |     where

28 | |         T: Iterator<Item = Pixel<Color>>,

29 | |     {

30 | |         self.draw_helper(WIDTH, HEIGHT, item_pixels);

31 | |     }

   | |_____^ impl has extra requirement `T: core::iter::Iterator`

error[E0276]: impl has stricter requirements than trait

  --> src/epd2in9/graphics.rs:26:5

   |

26 | /     fn draw<T>(&mut self, item_pixels: T)

27 | |     where

28 | |         T: Iterator<Item = Pixel<Color>>,

29 | |     {

30 | |         self.draw_helper(WIDTH, HEIGHT, item_pixels);

31 | |     }

   | |_____^ impl has extra requirement `T: core::iter::Iterator`
```

caemor:
Additionally you could update the `fn draw_helper<T>` in graphics as well as it should accept the same type. Also that should remove the need to add `.into_iter()` in all the implemenations

caemor:
:+1: 
Ignore the clippy failure for the outdated embedded hal traits. Now it's just the examples also using the outdated `with_strokes` and `with_fill` methods I believe

dbr:
Ahh, the examples were still specifying `embedded-graphics="0.4.5"`

Thanks for the pointers BTW, and the useful library!

examples/epd1in54_no_graphics/Cargo.toml
```diff
@@ -8,5 +8,5 @@ edition = "2018"
 
 epd-waveshare = { path = "../../", default-features = false, features = ["epd1in54"]}
 
-linux-embedded-hal = "0.2.2"
+linux-embedded-hal = "0.5.2"
```
caemor: copy-paste mistake most likely. this is an example without the embedded-graphics lib so nothing should be changed here ;-)

caemor:
Thanks :-)

PR #33, opened 2019-11-16
EPD7in5 support

str4d:


str4d:
This seems to be about twice as slow to write to the screen as the Waveshare Python controller, despite using the same panel settings.

str4d:
Measuring some drawing times (`DATA_START_TRANSMISSION_1` + `DISPLAY_REFRESH`):

- Old Python driver: 24s
  - Buffer is 30720 bytes (`= HEIGHT * WIDTH / 8`)
  - Writes 4 data bytes per buffer byte
- Latest Python driver: 30s
  - Buffer is `HEIGHT * WIDTH / 4` bytes
  - Writes 2 data bytes per buffer byte
- This Rust driver: 53.5s
  - Buffer is `HEIGHT * WIDTH / 8` bytes
  - Writes 4 data bytes per buffer byte

caemor:
> Measuring some drawing times (`DATA_START_TRANSMISSION_1` + `DISPLAY_REFRESH`):
> 
>     * Old Python driver: 24s
>       
>       * Buffer is 30720 bytes (`= HEIGHT * WIDTH / 8`)
>       * Writes 4 data bytes per buffer byte
> 
>     * Latest Python driver: 30s
>       
>       * Buffer is `HEIGHT * WIDTH / 4` bytes
>       * Writes 2 data bytes per buffer byte
> 
>     * This Rust driver: 53s
>       
>       * Buffer is `HEIGHT * WIDTH / 8` bytes
>       * Writes 4 data bytes per buffer byte

You are using one of the raspberry pis?
Can you give me a link to the specific version you were using? I don't know why we are so much slower besides maybe doing some additional checks, but that shouldn't slow us down so much :thinking:

str4d:
Added to the readme.

> Can you give me a link to the specific version you were using?

I'm using the [Raspberry Pi Zero W](https://www.raspberrypi.org/products/raspberry-pi-zero-w/).

caemor:
Added a new issue for the performance tracking

caemor:
@str4d do you remember if you used the same spi-frequency?

PR #35, opened 2019-12-13
Add support for the epd1in54b display

jkristell:
This RFC pull request adds support for the 2 color (red and black) display epd1in54b .

To support the two different planes of colored ink I have added a new trait called WaveshareTwoColorDisplay with on method - update_both_planes() for updating both planes at once. The old update_frame() just clears the color-layer when updating the black layer


caemor:
Looks great on a fast skim. I will make a full review next week, but could you in the meantime please add the new device to the Readme and add a short description/doc comments to the new trait?

src/traits.rs
```diff
@@ -52,6 +52,23 @@ where
     ) -> Result<(), SPI::Error>;
 }
```
caemor: Just add a short description what this trait is doing here

jkristell:
Added some documentation.

Perhaps the trait should be named WaveshareThreeColorDisplay instead as that's the terminology that Waveshare use.

caemor:
Then we should use the same terminology so it's easier for people to switch :-)
And can you please run cargo format, then it would be ready to be merged.
Greetings Chris

jkristell:
I'm not really happy about the ```update_both_planes()``` name. What do you think about renaming it to ```update_color_frame()```?

Naming is hard :)

```
 fn update_color_frame(
        &mut self,
        spi: &mut SPI,
        black: &[u8],
        color1: &[u8],
    ) -> Result<(), SPI::Error>;
```

jkristell:
Trait and method names updated, and code was formatted with ```cargo fmt```.

jkristell:
Graphics works. Yeah exactly like that. Two "displays". I have test code for this that I will clean up and publish as an example in my [nucleo-f401re](https://github.com/jkristell/nucleo-f401re) crate.

jkristell:
https://github.com/jkristell/nucleo-f401re/blob/add-eink-example/examples/eink.rs

caemor:
Thanks for your work :-)

PR #36, opened 2020-01-11
Add epd7in5_v2

asaaki:
When I ordered such 7.5" display and tried to use this library I was confused, as nothing worked. After some time I realized the very red note and thought »Oh, I must be a lucky one« since nobody besides Waveshare has support for V2 yet.

After spending hours in the python code and the spec PDF I came up with the following code, which works fine for my display. Not sure if anyone is willing to spend the money to confirm it. :sweat_smile: 

I also got rid of the deprecation warnings from embedded-hal, shouting at me to not use digital::v1 anymore. My workaround might look hacky, but should not be really worse than the logic before, though the `let _ = …` might be a bit unpleasing, as well as the `.unwrap_or(false)` — yes, I make probably a stupid assumption here to unlock the busy state on error, we could also negate and potentially lock forever; dunno what's best. Given that we ignored potentially failing state changes, I think it should be mostly fine.

---

Commit asaaki/epd-waveshare@19e3d51 was an attempt to fix the digital::v2 fallability, but I discovered that it would at least not work with linux-embedded-hal due to error type mismatches. And the examples step of CI is also pretty unhappy about it.

### DIFF
--- a/src/interface.rs
+++ b/src/interface.rs
@@ -151,7 +151,8 @@ where
     /// Most likely there was a mistake with the 2in9 busy connection
     /// //TODO: use the #cfg feature to make this compile the right way for the certain types
     pub(crate) fn is_busy(&self, is_busy_low: bool) -> bool {
-        (is_busy_low && self.busy.is_low().unwrap_or(false)) || (!is_busy_low && self.busy.is_high().unwrap_or(false))
+        (is_busy_low && self.busy.is_low().unwrap_or(false))
+            || (!is_busy_low && self.busy.is_high().unwrap_or(false))
     }
 
     /// Resets the device.
--- a/src/interface.rs
+++ b/src/interface.rs
@@ -2,7 +2,7 @@ use crate::traits::Command;
 use core::marker::PhantomData;
 use embedded_hal::{
     blocking::{delay::*, spi::Write},
-    digital::*,
+    digital::v2::*,
 };
 
 /// The Connection Interface of all (?) Waveshare EPD-Devices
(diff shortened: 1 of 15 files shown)


caemor:
> Commit asaaki/epd-waveshare@19e3d51 was an attempt to fix the digital::v2 fallability, but I discovered that it would at least not work with linux-embedded-hal due to error type mismatches. And the examples step of CI is also pretty unhappy about it.

In my first attempt to update to digital::v2 the result also wasn't what I wanted so I am currently waiting for the digital::v3/other results of the discussion (https://github.com/rust-embedded/wg/pull/393)

src/interface.rs
```diff
@@ -151,7 +151,8 @@ where
     /// Most likely there was a mistake with the 2in9 busy connection
     /// //TODO: use the #cfg feature to make this compile the right way for the certain types
     pub(crate) fn is_busy(&self, is_busy_low: bool) -> bool {
-        (is_busy_low && self.busy.is_low()) || (!is_busy_low && self.busy.is_high())
+        (is_busy_low && self.busy.is_low().unwrap_or(false))
+            || (!is_busy_low && self.busy.is_high().unwrap_or(false))
```
caemor: this gives different results for the devices with different is_busy_low variants. this should be changed to either return false (or true) for both cases. False might be better, because when we get an error we should better just block other stuff than maybe breaking things...

caemor:
Also thanks for your work :-) :+1:

asaaki:
Thanks for the pointer. Seems to be a very messy situation. Will undo all digital::* changes then.

asaaki:
> Would it make sense to add your update_and_display_frame to the main trait?

Could make sense, waveshare's python reference implementation uses a single `display` function for non-v2 as well, so providing a consolidated function in the main trait could be beneficial for everyone.
Will give it a try.

caemor:
> Will undo all digital::* changes then

I think we can leave your current v2 impl/changes and just keep in mind that we need to improve this situation later on (with v3) for which I made a new issue earlier.

caemor:
I am gonna merge this for now and make the change to the main trait in a seperate pr

asaaki:
Sorry, was occupied with other stuff. Thanks for taking it in. 🙇🏻‍♂️

PR #41, opened 2020-03-25
Add epd2in9b support

michaelbeaumont:
Any ideas what to call `TriColor::Red` so it makes sense for the yellow displays?

I'm using an stm32f042 so I get stack overflows if I try to allocate 2 pixel buffers at this resolution. Hence the separate `mono` and `red` update functions. Maybe they should be part of `ThreeColorDisplay`?

Ultimately I'd like to add `Drawing<TriColor>` support. Maybe there's a clever way to `draw` them without allocating 2 buffers. Right now I use `Black` as `Red` in `update_{color,red}_frame` like in epd1in54b.


src/color.rs
```diff
@@ -9,6 +9,17 @@ pub enum Color {
...
 }
 
+/// Only for the Black/White/Color-Displays
+#[derive(Clone, Copy, PartialEq, Debug)]
+pub enum TriColor {
+    /// Black color
+    Black,
+    /// White color
+    White,
+    /// Red color
```
caemor: Make it neutral to somehow combine yellow and red

src/epd2in9b/mod.rs
```diff
@@ -0,0 +1,322 @@
...
+pub(crate) mod command;
+use self::command::Command;
+
+#[cfg(feature = "graphics")]
+mod graphics;
+#[cfg(feature = "graphics")]
+pub use self::graphics::Display2in9b;
+
+
+/// EPD2in9b driver
```
caemor: Could you please add a simple example here?

src/epd2in9b/mod.rs
```diff
@@ -0,0 +1,322 @@
...
+        self.interface
+            .cmd(spi, Command::DATA_START_TRANSMISSION_1)?;
+        self.interface.data(spi, black)?;
+        Ok(())
+    }
+
+    /// Update only red data of the display.
+    ///
+    /// This data takes precedence over the black/white data.
+    pub fn update_red_frame(&mut self, spi: &mut SPI, red: &[u8]) -> Result<(), SPI::Error> {
```
caemor: rename it to update_color_frame or similar please and try adding it to the threecolordisplay

src/epd2in9b/mod.rs
```diff
@@ -0,0 +1,322 @@
...
+            spi,
+            Command::VCOM_AND_DATA_INTERVAL_SETTING,
+            &[border | VCOM_DATA_INTERVAL],
+        )
+    }
+
+    /// Update only the black/white data of the display.
+    ///
+    /// Finish by calling `update_red_frame`.
+    pub fn update_mono_frame(&mut self, spi: &mut SPI, black: &[u8]) -> Result<(), SPI::Error> {
```
caemor: Please also add this to the threecolor display

src/epd2in9b/graphics.rs
```diff
@@ -0,0 +1,45 @@
...
+impl Default for Display2in9b {
+    fn default() -> Self {
+        Display2in9b {
+            buffer: [DEFAULT_BACKGROUND_COLOR.get_byte_value(); NUM_DISPLAY_BITS as usize],
+            rotation: DisplayRotation::default(),
+        }
+    }
+}
+
+impl Drawing<Color> for Display2in9b {
```
caemor: I merged my updated branch yesterday, so you need to update this to the newest emedded-graphics version too.

caemor:
Since we are depending on the embedded-graphics library for most of the graphic stuff I would like to behave similar to it. Also the branch nidus is more uptodate then the master branch currently
- https://docs.rs/embedded-graphics/0.6.0/embedded_graphics/pixelcolor/index.html
- https://docs.rs/embedded-graphics/0.6.0/embedded_graphics/pixelcolor/trait.PixelColor.html
- https://docs.rs/embedded-graphics/0.6.0/embedded_graphics/pixelcolor/raw/index.html

> Any ideas what to call TriColor::Red so it makes sense for the yellow displays?
Maybe just call it Colorful (or preferably colored but according to my translator this is considered offensive in the usa nowadays) or Color. 

> I'm using an stm32f042 so I get stack overflows if I try to allocate 2 pixel buffers at this resolution. 
Did you try to use VarDisplay to update smaller parts of the display? 

> Hence the separate mono and red update functions. Maybe they should be part of ThreeColorDisplay?
It would make sense to add them to the ThreeColorDisplay-Trait yes. So you could just use one of them. 

> Ultimately I'd like to add Drawing<TriColor> support. Maybe there's a clever way to draw them without allocating 2 buffers. Right now I use Black as Red in update_{color,red}_frame like in epd1in54b.

See the links from above. If one could use the update_partial_frame it would also be possible to only update smaller parts of the screen. It would be great if one could read the current buffer directly on the epd so we wouldn't need to store it locally as well but sadly its not supported.

src/epd2in9b/mod.rs
```diff
@@ -0,0 +1,322 @@
...
+        let h = self.height();
+
+        self.command(spi, Command::RESOLUTION_SETTING)?;
+
+        self.send_data(spi, &[w as u8])?;
+        self.send_data(spi, &[(h >> 8) as u8])?;
+        self.send_data(spi, &[h as u8])
+    }
+
+    /// Set the outer border of the display to the chosen color.
```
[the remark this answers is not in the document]
michaelbeaumont: It looks like a rectangle of maybe 15 extra pixels around the screen, exactly.

src/epd2in9b/mod.rs
```diff
@@ -0,0 +1,322 @@
...
+        self.interface
+            .cmd(spi, Command::DATA_START_TRANSMISSION_2)?;
+        self.interface.data_x_times(spi, color, NUM_DISPLAY_BITS)?;
+
+        self.wait_until_idle();
+        Ok(())
+    }
+
+    #[allow(unused)]
+    fn update_partial_frame(
```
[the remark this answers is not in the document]
michaelbeaumont: It is possible but I just haven't tested it yet. I'd add support for that next

src/epd2in9b/mod.rs
michaelbeaumont: `update_color_frame` exists already for `ThreeColorDisplay` and updates both at the same time. 
I also find it a better name though for this case than for mono + color, could the existing one be renamed? Otherwise `update_only_color_frame`... naming things is hard

src/color.rs
michaelbeaumont: Hmm yeah you're right that `Colored` has certain negative connotations.
Going off the [wikipedia page for Colorfulness](https://en.wikipedia.org/wiki/Colorfulness) I suppose the most correct name would be `Chromatic` although `Colorful` is good too

caemor: Sounds good

src/epd2in9b/mod.rs
caemor: Yes this makes it more concise :+1:

src/epd2in9b/mod.rs
michaelbeaumont: done!

Issue #26, opened 2019-03-31
Support Color Displays?


jamesmunns:
I recently ordered a collection of the WaveShare displays, including:

* 4.2" Yellow/Black/White
* 4.2" Black/White
* 4.2" Red/Black/White
* 7.5" Black/White
* 5.83" Red/Black White

I'm happy to contribute or help test, as I plan to use these on a couple of embedded devices for my [internet of streams](https://github.com/ferrous-systems/internet-of-streams) project. I was wondering if you had any notes already on where I should be looking to help add support for the color displays.

caemor:
Sorry for the late response. 

- the 4.2" B/W is working already. 
- the 7.5" B/W should also be pretty easy to add (I also got one at home but didn#t have the time to add it and test it)

AFAIK the colored ones just use two buffers instead of one so that also shouldn't be too complicated too add (where the biggest problem will be how to add/change [color.rs](https://github.com/caemor/epd-waveshare/blob/master/src/color.rs) and if 2 small buffers are better or one big buffer). I am really unsure about this problem.

Question to the naming convention: the call it 4in2(b) but wouldn't 4in2_red or something similar be better?

### Steps that need to be done to add a display generally: (e.g for the 4in2 (b)/red one)

1. Get the **Command table** from the datasheet (e.g here: https://www.waveshare.com/wiki/4.2inch_e-Paper_Module_(B)#Datasheets ) and add it to a command enum like here (https://github.com/caemor/epd-waveshare/blob/master/src/epd4in2/command.rs) in a epd4in2b folder. This can also be found in waveshares new github repo (e.g.: https://github.com/waveshare/e-Paper/blob/master/4.2inch_e-paper_b%26c_code/wiringpi/obj/EPD_4in2b.h but be aware of quite a lot of copy&paste errors like the wrong ifndef epd2in7b at the top and better crosscheck it once more)

2. Optional: Add the **LUT constants** (like https://github.com/caemor/epd-waveshare/blob/master/src/epd4in2/constants.rs) if you can find any. But since the color ones only allow the default refresh they are most likely not included in the repos. This is also not too important/optional.

3. Add following constants from the next four lines to epd4in2b/mod.rs (https://github.com/caemor/epd-waveshare/blob/master/src/epd4in2/mod.rs#L61): width, height, default_background_color (twice for red/black displays?), is_busy_low

4. Implement **WaveshareDisplay** and **InternalWiAdditions** trait for your EPD4in2_Red struct.
[...]

PR #45, opened 2020-04-14
WIP: Add 7in5bc support, #26

marcelbuesing:
I'm working on 7in5bc support for aer and since I wanted to try out the colors I added the 7in5bc version. It's mostly copy and paste besides docs and the `fn update_color_frame`. It looks like this is a little different to e.g. the 2in9bc where the update of chromatic and monochromatic part appears to happen separately. 
I still added `update_achromatic_frame` and `update_chromatic_frame` but it adds little value since it's not possible to set e.g. the chromatic part first and then only update the achromatic part (haven't tested if it works the other way around).

image

### DIFF
--- a/src/epd7in5bc/mod.rs
+++ b/src/epd7in5bc/mod.rs
@@ -111,7 +111,7 @@ where
         delay: &mut DELAY,
     ) -> Result<(), SPI::Error> {
         // Reset the device
-        self.interface.reset(delay);
+        self.interface.reset(delay, 10);
 
         // Set the power settings
         self.cmd_with_data(spi, Command::POWER_SETTING, &[0x37, 0x00])?;
--- a/src/epd7in5bc/mod.rs
+++ b/src/epd7in5bc/mod.rs
@@ -1,0 +1,430 @@
+//! A simple Driver for the Waveshare 7.5" (B/C) E-Ink Display via SPI
+//!
+//! # References
+//!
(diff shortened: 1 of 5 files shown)


caemor:
> I still added update_achromatic_frame and update_chromatic_frame but it adds little value since it's not possible to set e.g. the chromatic part first and then only update the achromatic part (haven't tested if it works the other way around).

Could you please add this to the documentation of this display and to the these two functions. 
This might be useful for people with smaller RAMs to first update the achromatic and then reuse the same buffer for the chromatic part.

Everything else looks really good

caemor (changes requested):
See last message and could you maybe rebase this to main?

src/epd7in5bc/mod.rs
```diff
@@ -0,0 +1,430 @@
...
+    DC: OutputPin,
+    RST: OutputPin,
+{
+    fn init<DELAY: DelayMs<u8>>(
+        &mut self,
+        spi: &mut SPI,
+        delay: &mut DELAY,
+    ) -> Result<(), SPI::Error> {
+        // Reset the device
+        self.interface.reset(delay, 10);
```
marcelbuesing, line 114: Assuming 10 is fine here but I can not verify this due to missing display.

marcelbuesing:
> See last message and could you maybe rebase this to main?

Sorry my 7in5c display died some time last year when I accidentally dropped it and it did not make sense to order the v1 version again. I rebased the branch but it would be good to try the chromatic/achromatic fns as maybe I just tried it the wrong way around last year.

PR #48, opened 2020-05-16
EPD4in2: Don't set the resolution (and some more) over and over again.

mgottschlag:
If the resolution ever changes, something has horribly gone wrong.


caemor:
This is necessary since `update_partial_frame` changes the resolution and if i remember correctly it was rather cheap (fast) in comparison to updating the full frame so I chose the solution where nothing can go wrong when forgetting to reset it.

So if we want to reduce setting the resolution each time we would either need some seperate function to reset it or, alternatively change the description of `update_partial_frame` and add a note there that a full init is necessary when wanting to use the update(full)_frame once more.

PR #52, opened 2020-09-21
Better handling for screen size not divisible by 8

dkm:
Screen with width not divisible by 8 require special care when allocating the
buffer and when handling rotation.

Define a function for getting the needed buffer size (round up to next byte).
Change how rotation is done by changing coordinates instead of direct pixel
access.

### DIFF
--- a/src/graphics.rs
+++ b/src/graphics.rs
@@ -1,5 +1,6 @@
 //! Graphics Support for EPDs
 
+use crate::buffer_len;
 use crate::color::Color;
 use embedded_graphics::{pixelcolor::BinaryColor, prelude::*};
 
@@ -88,6 +89,7 @@ pub trait Display: DrawTarget<BinaryColor> {
 ///
 /// The buffer can be created as following:
 /// buffer: [DEFAULT_BACKGROUND_COLOR.get_byte_value(); WIDTH / 8 * HEIGHT]
+/// If WIDTH is not a multiple of 8, don't forget to round it up (ie. (WIDTH + 7) / 8)
 ///
 /// Example:
 /// ```rust,no_run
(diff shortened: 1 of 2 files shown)


caemor:
Screen with width not divisible by 8 require special care when allocating the
buffer and when handling rotation.

> Define a macro for getting the needed buffer size (round up to next byte).
Is there a reason why you used a macro instead of a normal function here?

> Change how rotation is done by changing coordinates instead of direct pixel
access.
Looks much simpler now :+1: 

Else I think it looks ready to accept.

dkm:
> > Define a macro for getting the needed buffer size (round up to next byte).

> Is there a reason why you used a macro instead of a normal function here?

I guess a bad habit of doing too much C. I'll try using a fn instead ! I'm still not familiar with what can be used and where.

caemor:
> > > Define a macro for getting the needed buffer size (round up to next byte).
> 
> > Is there a reason why you used a macro instead of a normal function here?
> 
> I guess a bad habit of doing too much C. I'll try using a fn instead ! I'm still not familiar with what can be used and where.

Yeah I just prefer functions since they are easier to debug and the compiler can check them more easily as well. But since macros can to more than functions they are sometimes just needed :-D

dkm:
I've moved the function in the top module to be able to use it even when graphics is not used.
I also tried to change all `WIDTH/8 * HEIGHT` everywhere but realized it was maybe not a good idea and reverted it. Let me know if you want me to change all screen's code in this way.

src/graphics.rs
```diff
@@ -120,10 +122,10 @@ pub struct VarDisplay<'a> {
 impl<'a> VarDisplay<'a> {
     /// Create a new variable sized display.
     ///
-    /// Buffersize must be at least width / 8 * height bytes.
+    /// Buffersize must be at least (width + 7) / 8 * height bytes.
```
caemor, line 125: Maybe add a small rounded up in brackets behind the line to make it easier to understand why the + 7 is necessary on a quick glance.

/// Buffersize must be at least (width + 7) / 8 * height bytes. (rounded up!)

caemor (approved):

caemor:
I think I am just gonna merge this now. Looks great :+1: 

> I also tried to change all WIDTH/8 * HEIGHT everywhere but realized it was maybe not a good idea and reverted it. Let me know if you want me to change all screen's code in this way.

I added this to my todo list. But lets first merge this now, so we can get ready to merge #51 as well

dkm:
Great, thanks !

PR #53, opened 2020-10-05
Add 2in13 to the list of supported screens

dkm:
Add 2in13 to the table of supported device


caemor:
Could you please rename it to v2? 
Thanks in advance

dkm:
Done !

Issue #47, opened 2020-05-16
4.2": Quick refresh and background color


mgottschlag:
I have experimented with the 4.2 inch e-ink display from waveshare, and the last days I have tried to make quick refresh work. I have observed a number of deficiencies in this library, some of which I do not know how to fix:

- [ ] Newer 4.2" B/W displays seem to require different LUTs for quick refresh.

    As indicated by a comment below [Ben Krasnow's blog post on quick refresh on this display](https://benkrasnow.blogspot.com/2017/10/fast-partial-refresh-on-42-e-paper.html), there are two versions of the display. I copied the LUTs from the example code for the greyscale Good Display display (does that mean that my B/W display properly supports greyscale? Yay!) and those result in proper black/white pixels, whereas the LUTs in this library currently only result in grey pixels.

    Link to the example code: http://www.e-paper-display.com/download_detail/downloadsId=1022.html

    My commit (the old LUTs should probably be accessible though - preferrably not behind a feature gate, but selectable at runtime): https://github.com/mgottschlag/epd-waveshare/commit/a24bec5be720c89cedda513a139bc983b0c5e1d0

- [ ] I am unsure what the point of the "background color" provided by this library is - the documentation is lacking and I think usage the background color as-is is wrong and unusable.

    My understanding of the background (or "old" frame as called by the documentation is that it allows to choose different LUTs based on what the previous image was. Note that the data is actually required for quick refresh using the LUTs in the example code above, as those completely skip pixels that do not change color!

    Also, using the wrong background data might cause burn-in issues. According to the internet (I think Ben Krasnows youtube video explained something like this), the applied voltage needs to be on average 0 to prevent charging the display, which causes burn-in. I don't understand the format of the LUTs, but if they are not completely balanced (and inofficial quick refresh LUTs likely are not!), then using the same LUT over and over again causes the display to be charged (whereas only applying voltages during black/white and white/black transitions might cause the effects to cancel each other out and also greatly reduces the number of pixel changes).

    At least, the library requires updates to the documentation.

- [ ] The library does not provide a method to upload the "old" ("background") data.

    The official quick refresh LUTs for this display depend on having the old data.

    My take on this issue: https://github.com/mgottschlag/epd-waveshare/commit/dbd065a71595b62147b5a4af8846f1ce1af6e191

mgottschlag:
[...]

PR #54, opened 2020-10-06
Added partial/quick refresh support

David-OConnor:
https://github.com/caemor/epd-waveshare/issues/47

@mgottschlag


src/epd4in2/constants.rs
```diff
@@ -1,39 +1,39 @@
...
-0x00, 0x0A, 0x01, 0x00, 0x00, 0x01,        
-0x00, 0x0E, 0x0E, 0x00, 0x00, 0x02,        
-0x00, 0x00, 0x00, 0x00, 0x00, 0x00,        
-0x00, 0x00, 0x00, 0x00, 0x00, 0x00,        
+0x00, 0x08, 0x00, 0x00, 0x00, 0x02,
+0x60, 0x28, 0x28, 0x00, 0x00, 0x01,
+0x00, 0x14, 0x00, 0x00, 0x00, 0x01,
+0x00, 0x12, 0x12, 0x00, 0x00, 0x01,
+0x00, 0x00, 0x00, 0x00, 0x00, 0x00,
+0x00, 0x00, 0x00, 0x00, 0x00, 0x00,
```
caemor, line 8: These are the offical LUTs from Waveshare (see the full review text for more infos)

caemor (changes requested):
The current normal LUTs are the official LUTs from waveshare. The fast ones are from Krasnow.
I don't want to change them, since they are working well for the old version of the display (simple b/w) (https://github.com/waveshare/e-Paper/blob/751a9fb93fdd486511222777b0070c51bf436386/Arduino/epd4in2/epd4in2.cpp#L456). But as you can see further below they have different LUTS for the greyscale version as well. 

So I would prefer to add seperate LUTs for this and maybe also make it a seperate display (v2). Since its not a simple B/W Display anymore but supports B, G1, G2 and W.

The c version of the new and old LUTS from above:
https://github.com/waveshare/e-Paper/blob/master/STM32/STM32-F103ZET6/User/e-Paper/EPD_4in2.c

src/epd4in2/constants.rs
caemor: You can still add the new ones but let the old one there as well.
You have tested the new ones on the 4gray display or also on others?

David-OConnor: I don't have a 4gray display to test on. This is the first I've heard of them.

caemor: with 4gray I mean the one that has black, white, grey1 and grey2? I thought thats also the one @mgottschlag has.
I though that is basically the new variant of the epd4in2.

caemor: > As indicated by a comment below Ben Krasnow's blog post on quick refresh on this display, there are two versions of the display. I copied the LUTs from the example code for the greyscale Good Display display (does that mean that my B/W display properly supports greyscale? Yay!) and those result in proper black/white pixels, whereas the LUTs in this library currently only result in grey pixels.

David-OConnor: I'm still not sure about the different versions. Going to rewatch that vid. I haven't seen a distinction or any advertisement of "4 gray shades", like you do with the 3 color ones.

I'm happy to punt the LUT and version business to a new PR, and leave this one about quick refreshes. Or if we solve this, do it here.

David-OConnor: Both LUTs seem to work on the display I have.

caemor: Sorry for the delay.

Two things are currently missing to finish this:
- Can you maybe add the new LUT behind a new feature gate similar to the type_a_alternative_faster_lut feature?
- Make a seperate trait for the quick refresh features and add a function there which does both

src/epd1in54b/mod.rs
```diff
@@ -304,6 +306,48 @@ where
...
+    fn clear_partial_frame(
+        &mut self,
+        _spi: &mut SPI,
+        _x: u32,
+        _y: u32,
+        _width: u32,
+        _height: u32,
+    ) -> Result<(), SPI::Error> {
+        unimplemented!("Not yet implemented");
+    }
```
caemor, line 350: See other trait comment

src/epd4in2/constants.rs
```diff
@@ -1,6 +1,7 @@
 #[rustfmt::skip]
 pub(crate) const LUT_VCOM0: [u8; 44] = [
-0x00, 0x17, 0x00, 0x00, 0x00, 0x02,        
+// 0x00, 0x17, 0x00, 0x00, 0x00, 0x02,        
+0x40, 0x17, 0x00, 0x00, 0x00, 0x02,
```
caemor, line 4: Is there a reason for this line change? If I remember correctly the first byte is for the amount of repetitions or length of the following line and it was 0x00 elsewhere in the referenced stuff I believe

src/epd4in2/mod.rs
```diff
@@ -230,27 +230,17 @@ where
...
-        self.send_data(spi, &[(y >> 8) as u8])?;
-        self.send_data(spi, &[y as u8])?;
-
-        self.send_data(spi, &[((y + height - 1) >> 8) as u8])?;
-        self.send_data(spi, &[(y + height - 1) as u8])?;
+        self.interface.cmd(spi, Command::PARTIAL_IN)?;
+        self.interface.cmd(spi, Command::PARTIAL_WINDOW)?;
 
-        self.send_data(spi, &[0x01])?; // Gates scan both inside and outside of the partial window. (default)
+        self.shift_display(spi, x, y, width, height)?;
```
caemor, line 243: shift_display is a great improvement!

caemor (changes requested):
See the other comments

src/epd4in2/constants.rs
David-OConnor: Sep trait added

src/epd1in54/mod.rs
```diff
@@ -282,6 +284,48 @@ where
...
+    fn clear_partial_frame(
+        &mut self,
+        _spi: &mut SPI,
+        _x: u32,
+        _y: u32,
+        _width: u32,
+        _height: u32,
+    ) -> Result<(), SPI::Error> {
+        unimplemented!("Not yet implemented");
+    }
```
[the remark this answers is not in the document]
David-OConnor: Done

src/epd4in2/constants.rs
David-OConnor: I noticed it in the video referenced above. No idea what the truth data is! Thoughts?

src/epd4in2/mod.rs
David-OConnor: Thanks. It saved repetition with the things I added.

src/epd4in2/constants.rs
caemor: After looking through multiple sources and Krasnows video once more I think it should be 0x00 there since all the other LUTs for VCOM0 also used only that voltage source and since it doesn't make a difference (or does it?) I would leave it the old way and maybe just add a note there if someone is ever interested in testing this extensively.

caemor:
Other than that small thing with the LUT it looks ready to merge from my side.

Another option would be to bring in your alternative LUT from further at the top behind a feature gate.

mgottschlag:
I think the doc comment should
- [ ] state when the user needs to send the old data
- [ ] explain why this is necessary
- [ ] maybe give an example?

I think this whole old status/new status stuff is fairly confusing. I can write a patch for that this weekend if you want.

caemor:
> I think the doc comment should
> 
>     * [ ]  state when the user needs to send the old data
> 
>     * [ ]  explain why this is necessary
> 
>     * [ ]  maybe give an example?
> 
> 
> I think this whole old status/new status stuff is fairly confusing. I can write a patch for that this weekend if you want.

Sure would be great

caemor:
@David-OConnor Sorry, for the late replay! It looks great now. We just need to make it so the build won't fail anymore when going through the tests. Currently the testcase in `src/traits.rs - traits::QuickRefresh (line 244) ... FAILED`

PR #55, opened 2020-10-17
make reset timing device specific

icewind1991:
reset timings seem to be device specific, trying to reset the 7in5_v2 with timings from other devices does not reset the display.

Timings taken from https://github.com/waveshare/e-Paper/blob/master/RaspberryPi%26JetsonNano/python/lib/waveshare_epd

### DIFF
--- a/src/epd7in5_v2/mod.rs
+++ b/src/epd7in5_v2/mod.rs
@@ -59,7 +59,7 @@ where
         delay: &mut DELAY,
     ) -> Result<(), SPI::Error> {
         // Reset the device
-        self.interface.reset(delay, 4);
+        self.interface.reset(delay, 2);
 
         // V2 procedure as described here:
         // https://github.com/waveshare/e-Paper/blob/master/RaspberryPi%26JetsonNano/python/lib/waveshare_epd/epd7in5bc_V2.py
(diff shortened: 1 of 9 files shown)


src/epd7in5_v2/mod.rs
```diff
@@ -59,7 +59,7 @@ where
         delay: &mut DELAY,
     ) -> Result<(), SPI::Error> {
         // Reset the device
-        self.interface.reset(delay);
+        self.interface.reset(delay, 4);
```
caemor, line 62: https://github.com/waveshare/e-Paper/blob/751a9fb93fdd486511222777b0070c51bf436386/RaspberryPi%26JetsonNano/python/lib/waveshare_epd/epd7in5_V2.py#L52

I think it should be only 2 ms for this display. You might have looked for the B/C variants (red/yellow) which have 4 ms.

caemor (changes requested):
Thanks for the pr!

As can be seen by the todos there, the reset timing could be improved even further by extensive testing. I think I got them down to a much lower limit in my personal tests but wasn't sure if this would be stable everywhere. 

So your changes look like a good improvement in the right direction :+1: 
Although you might have taken the wrong shorter timings for the epdi7in5v2 (see the comment there :D)

icewind1991:
You're right, I was probably looking at the wrong variant, confirmed that 2ms also works correctly

caemor (approved):

caemor:
Is it ready to merge?

icewind1991:
ready from my pov

caemor:
Thanks for you work!

PR #59, opened 2020-11-16
Add Support for 5.65 Inch 7 Color (F) Display

crzysdrs:


src/graphics.rs
```diff
@@ -186,30 +245,44 @@ fn outside_display(p: Point, width: u32, height: u32, rotation: DisplayRotation)
...
+
+#[rustfmt::skip]
+//returns index position in the u8-slice and the bit-position inside that u8
+fn find_oct_position(x: u32, y: u32, width: u32, height: u32, rotation: DisplayRotation) -> (u32, bool) {
+    let (nx, ny) = find_rotation(x, y, width, height, rotation);
+    (
+        nx / 2 + (width / 2) * ny,
+        (nx & 0x1) == 0,
+    )
+}
```
caemor, line 280: Maybe add a small description for the true, false case here

src/graphics.rs
```diff
@@ -85,6 +85,65 @@ pub trait Display: DrawTarget<BinaryColor> {
...
+/// - Drawing (With the help of DrawTarget/Embedded Graphics)
+/// - Rotations
+/// - Clearing
+pub trait OctDisplay: DrawTarget<OctColor> {
+    /// Clears the buffer of the display with the chosen background color
+    fn clear_buffer(&mut self, background_color: Color) {
+        for elem in self.get_mut_buffer().iter_mut() {
+            *elem = background_color.get_byte_value();
+        }
+    }
```
caemor, line 100: could you try if the clear buffer function could be added to the default trait implemenation? 
Since it seems to be the same function anyway

src/epd5in65f/graphics.rs
```diff
@@ -0,0 +1,207 @@
...
+            assert_eq!(
+                OctColor::split_byte(byte),
+                Ok((
+                    epd5in65f::DEFAULT_BACKGROUND_COLOR,
+                    epd5in65f::DEFAULT_BACKGROUND_COLOR
+                ))
+            );
+        }
+    }
+}
```
caemor, line 207: Could you please  add some simple check for the colors in the buffer here so we have tests whenever we want to change anything around the general octocolor parts?

Cargo.toml
```diff
@@ -9,7 +9,7 @@ license = "ISC"
 name = "epd-waveshare"
 readme = "README.md"
 repository = "https://github.com/Caemor/epd-waveshare.git"
-version = "0.4.0"
+version = "0.5.0"
```
caemor, line 12: Do you need it released soon on crates.io? 
Else I would most likely wait for the embedded-hal 1.0 release, then fix/include a few of the nearly finished prs and also release a 1.0 version. But I can always release another version inbetween if needed earlier.

caemor (changes requested):
Thanks for your PR. Looks really good, there are just some small things I'd like to see changed (see the code comments)

Have you tried the display driver already? Do you have some pictures and how long does it take you for a refresh?

crzysdrs:
Example image showing all colors. Note the leftmost greyish color is HiZ, and is not formally defined as any particular color. It takes about ~8s to draw refresh the display. It's worth noting that the docs recommend drawing HiZ before each image to prevent the eInk ghosting, so for better imaging it takes double the time.
image

src/epd5in65f/graphics.rs
crzysdrs: Added some tests.

src/graphics.rs
crzysdrs: This was an oversight, I have updated the method.

src/graphics.rs
crzysdrs: Added some comments.

Cargo.toml
crzysdrs: Nope, take your time.

caemor (approved):

caemor:
Thanks! This looks great :+1:

PR #60, opened 2021-01-01
Add support for epd2in7b

pjsier:
Adds support for EPD 2.7" B Tri-Color display based on the instructions in #26. I based most of this off of the existing implementations, but let me know if I missed anything!

One quirk of this is that the Waveshare implementation [flips the display buffer data that it sends](https://github.com/waveshare/e-Paper/blob/66ff76f87ecac4f7ef338347695481ca14b58db1/RaspberryPi_JetsonNano/python/lib/waveshare_epd/epd2in7b.py#L232-L244). I didn't see this in any of the other implementations, but it only works for me when I do the same, so I added that in the `send_buffer_helper` function.

I didn't implement anything for the `TriColor` enum since it looks like there wasn't much there, but using `Color::Black` in any of the `chromatic` functions allows for drawing red. Quick example of drawing a red box with a partial update here:

```rust
let small_buf = [Color::Black.get_byte_value(); 32];
epd2in7b.update_partial_chromatic_frame(&mut spi, &small_buf, 80, 80, 16, 16)?;
epd2in7b
    .display_partial_frame(&mut spi, 80, 80, 16, 16)
    .expect("partial displaying");
```

I also wasn't able to find quick update partial look-up tables, so I've left that part out.


caemor (changes requested):
Thanks for you PR. 

Need to take a better look at it tomorrow but from a quick glance it looks good.
Can you add the display to the readme as well?

pjsier:
No rush, thanks for taking a look! I just added it to the readme

caemor:
Thanks for your PR!

PR #62, opened 2021-01-21
Added QuickRefresh trait, and implemented it for 4.2 display

David-OConnor:
Clean rebase of [#54](https://github.com/caemor/epd-waveshare/pull/54). The final blocker on that one was Travis failing an example due to it not having an import. Now passes.


David-OConnor:
I have no idea what these new failures mean. Do we need to pass them to merge?

caemor:
Just codecov which is okay to fail. I should adjust that in the settings :+1:

caemor (approved):

src/traits.rs
```diff
@@ -229,3 +229,59 @@ where
...
 }
+
+/// Allows quick refresh support for displays that support it; lets you send both
+/// old and new frame data to support this.
+///
+/// When using the quick refresh look-up table, the display must receive separate display
+/// buffer data marked as old, and new. This is used to determine which pixels need to change,
+/// and how they will change. This isn't required when using full refreshes.
+///
+/// (todo: Example ommitted due to CI failures.)
```
caemor, line 240: What was the issue here?

We could fix this either with no_run or ignore, depending on what was missing (preferable no_run since then it still needs to compile and would make it better to recognize later regressions)
https://stackoverflow.com/questions/42348075/how-to-put-a-line-into-the-documentation-which-is-ignored-for-doc-tests#42348240

David-OConnor: It was this example:
```rust
/// Example:
///     epd.update_partial_old_frame(spi, disp.buffer(), x, y, frame_width, frame_height)
///         .ok();
///
///     disp.clear_buffer(Color::White);
///     // Execute drawing commands here.
///
///     epd.update_partial_new_frame(spi, disp.buffer(), x, y, frame_width, frame_height)
///         .ok();
```
It didn't like how `Color` wasn't imported explicitly.

PR #67, opened 2021-03-31
Added support for the 7.5 inch HD display

whiite:
I based this on the `epd7in5_v2` code but uses the hex codes from the 7.5" HD data sheet and ported the procedure from [epd7in5_HD.py](https://github.com/waveshare/e-Paper/blob/master/RaspberryPi_JetsonNano/python/lib/waveshare_epd/epd7in5_HD.py).

A difference with this display seems to be that the colours are inverted compared with the other displays (On/0xFF is white) so I've set `DEFAULT_BACKGROUND_COLOR` to `Black`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 ### Added
  - Added QuickRefresh Trait and implemented it for EPD4in2 in #62 (thanks to @David-OConnor)
  - Added Epd 2in7 (B) support in #60 (thanks to @pjsier)
+ - Added Epd 7in5 HD support
 
 ### Changed
  - Use specific ParseColorError instead of ()
--- a/README.md
+++ b/README.md
@@ -38,6 +38,7 @@ epd.update_and_display_frame(&mut spi, &display.buffer())?;
 
 | Device (with Link) | Colors | Flexible Display | Partial Refresh | Supported | Tested |
 | :---: | --- | :---: | :---: | :---: | :---: |
+| [7.5 Inch B/W HD (A)](https://www.waveshare.com/product/displays/e-paper/epaper-1/7.5inch-hd-e-paper-hat.htm) | Black, White | ✕ | ✕ | ✔ | ✔ |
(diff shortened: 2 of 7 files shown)


caemor:
Thanks for the PR. 

I will take a closer look next week.

caemor:
Could you please add it to the Readme and the Changelog (to Unreleased)?

And for the color type I think it would be better to either make a seperate Color type (e.g. ColorInverse) or explicitly add a comment to the docs of the new display for the color difference. I'm leaning towards a seperate type but both would be fine.
I believe the current status could be confusing for users who get different results for the same color type for different displays.

caemor (changes requested):
See comment above

whiite:
I've updated the changelog and readme to include the driver.

About the colour, I think that a new type would likely be the clearest solution however when I started adding the type I noticed that the current type is actually the correct type for the driver where a bit of value `1` is white while a bit of `0` is black and a byte of `0xFF` is used to set the screen to white which matches up with the `Color::White` enum. This driver is still inverted in comparison to the V2 driver which uses a byte of `0x00` to set the screen to white so I reckon the inverted color type should probably be applied to that driver instead. Should I open a separate PR for sorting that out?

In the mean time I've set the HD driver default background color to `Color::White` and add a doc comment about how the color is inverted when compared to the V2 driver.

caemor:
> Should I open a separate PR for sorting that out?
Yes. that would be great!

And thanks for your work!

caemor (approved):

caemor:
Thanks for the patches and sorry for the additional work. I finally switched to the recommended naming convention this week, which is why you needed to do all the additional work :see_no_evil:

caemor:
If you think it is ready I would merge.

whiite:
No problem! 🙂 I'm still relatively new to Rust and so wasn't sure how to check my changes passed all the GitHub checks locally. For future reference is there a good way to check it passes everything before I push?
Otherwise I think it's ready to be merged!

caemor:
The best thing is to run everything from https://github.com/caemor/epd-waveshare/blob/main/.github/workflows/rust.yml#L26 onwards yourself. (format, check & test are the important ones)

The clippy one should be allowed to fail but I didn't find a good way for it to run everything but have it still indicating that the clippy test failed similar to travis ci where I could set one of them to `allow-failure` so it would show a yellow success icon instead of a green one.

Issue #68, opened 2021-04-03
3.7in Display Problems


caemor:
Hey @caemor @jamesmunns ,
Thanks for pointers. I'm still figuring it out. I have some weird issues when I try to test the 3.7 display using a NRF52 board. Things I've checked so far - initialization and display sequence match C/Python examples, tried different LUT provided in those examples, width/height match as well, rotation only affects that tiny area where "circle" is shown.
Have you encountered this kind of effect?
[an attempt to draw a circle](https://user-images.githubusercontent.com/3869268/113447971-9628c880-9403-11eb-8a2d-e719fa8926ee.jpeg)

I got my hands on RPi and tested waveshare Python snippet with the same display, everything worked as expected. Digging into my implementation.

_Originally posted by @Alexx-G in https://github.com/caemor/epd-waveshare/issues/26#issuecomment-812680848_

caemor:
**Copied from #26**
> Thanks a lot @caemor for your effort!
> Is 3.7in b/w display supported?

Not yet. 
> If no, is #35 still an accurate example of adding a new display?

#35 is is a red/black/white display so it is slightly different. 
I believe the instructions from above might be more helpful but #35 as well as any of the other recent PRs adding a new display can help you, just be aware that your display differs in the colors. 

Since your 3in7 b/w display seems to have some gray tones you need to define a new color type in color.rs similar to what is defined in their wiki: https://www.waveshare.com/wiki/3.7inch_e-Paper_HAT 

So while finishing this I saw you already added a new comment :-D

> Thanks @jamesmunns !
I've started working on a PR yesterday, however I'm struggling finding a command table for 3.7 - https://www.waveshare.com/wiki/3.7inch_e-Paper_HAT

If you click on ressources you can find the specifications. (e.g. Command table starting from page 14 onwards https://www.waveshare.com/w/upload/7/71/3.7inch_e-Paper_Specification.pdf) 

And just feel free to make a WIP-PR and asks questions if you are stuck anywhere.

caemor:
> Thanks for pointers. I'm still figuring it out. I have some weird issues when I try to test the 3.7 display using a NRF52 board. > Things I've checked so far - initialization and display sequence match C/Python examples, tried different LUT provided in those examples, width/height match as well, rotation only affects that tiny area where "circle" is shown.
Have you encountered this kind of effect?

[...]

PR #69, opened 2021-04-03
WIP: Add support for EPD 3in7

Alexx-G:
Relates to #68 

I've relied on driver implementation for [STM32](https://github.com/waveshare/e-Paper/blob/master/STM32/STM32-F103ZET6/User/e-Paper/EPD_3in7.c) and [RPi](https://github.com/waveshare/e-Paper/blob/master/RaspberryPi_JetsonNano/python/lib/waveshare_epd/epd3in7.py).


caemor:
Hey Alexx, 
how is your current status?
greetings,
Chris

mangelajo:
I am preparing an updated PR for this display, as I recently got one looking for a fast-refresh display.

mangelajo:
> I am preparing an updated PR for this display, as I recently got one looking for a fast-refresh display.

Nice I got it working with the latest release. One of the issues seemed to be that the busy signal reading was inverted.

I have also verified that the Box::new(Displayxxx:default()) approach works, I'll document that in a separate PR too.

@caemor I guess that based on previous discussions the approach for 4-gray/1-gray is having 2 different versions of the display driver, right?

peckpeck:
Is it of any use having a B/W only version ? Is it any faster ?
If no a simple 4 gray implementation should be enough

mangelajo:
I have posted the PR here https://github.com/caemor/epd-waveshare/pull/129

We could close this PR at this point, original authorship is retained in the commit

mangelajo:
> Is it of any use having a B/W only version ? Is it any faster ? If no a simple 4 gray implementation should be enough

Memory usage is half, for 4 grayscale it doesn't work in ESP32C3 as far as I checked.

I will do timing checks , and try to factor-out common parts across 4gray/BW.

peckpeck:
i think esp32c3 has enough memory, but i understand the argument

PR #71, opened 2021-04-04
Partially fix #70

### LINKED ISSUE #70: 7.5" V2 display driver does not work
A few things I found while trying to get my display to work as expected:
* `epd7in5_v2v::EPD7in5::new(...)` hangs [here](https://github.com/caemor/epd-waveshare/blob/main/src/epd7in5_v2/mod.rs#L72) since the `wait_until_idle` implementation is wrong. In the [example C code provided by Waveshare](https://github.com/waveshare/e-Paper/blob/7c368a22afb5456fb4622c617e792bd2f83e878b/RaspberryPi_JetsonNano/c/lib/e-Paper/EPD_7in5_V2.c#L78) one can see that the driver sends a `GET_STATUS` command and waits 20ms.  This seems to be needed otherwise the busy pin wont update which is the root cause to the problem.

* Black and White seems to have been mixed up. Setting the fill-color to black makes the display white and vice versa. Same thing goes for text colors/background.

**A few things that might be important:**
* Cross-compiling from Arch Linux x64_64 to aarch64 with `aarch64-linux-gnu-gcc` targeting `aarch64-unknown-linux-musl`.
* Using the [rppal](https://github.com/golemparts/rppal) crate with the hal-unproven feature.
* Running on Raspberry Pi OS 64-bit Beta (Kernel 5.4.51-v8+) on a Raspberry Pi 3b+.

A hacked together version of epd-waveshare can be found [here](https://github.com/olback/epd-waveshare/tree/7.5hack) and an example can be found [here](https://github.com/olback/smart-home/blob/epd-working/portal/src/main.rs).

olback:
This is unfortunately a breaking change but it makes it easier to add more displays with similar requirements to the 7.5" V2 display.

### DIFF
--- a/src/epd7in5_v2/mod.rs
+++ b/src/epd7in5_v2/mod.rs
@@ -37,27 +37,24 @@ const IS_BUSY_LOW: bool = true;
 
 /// EPD7in5 (V2) driver
 ///
-pub struct EPD7in5<SPI, CS, BUSY, DC, RST> {
+pub struct EPD7in5<SPI, CS, BUSY, DC, RST, DELAY> {
     /// Connection Interface
-    interface: DisplayInterface<SPI, CS, BUSY, DC, RST>,
+    interface: DisplayInterface<SPI, CS, BUSY, DC, RST, DELAY>,
     /// Background Color
     color: Color,
 }
 
-impl<SPI, CS, BUSY, DC, RST> InternalWiAdditions<SPI, CS, BUSY, DC, RST>
-    for EPD7in5<SPI, CS, BUSY, DC, RST>
+impl<SPI, CS, BUSY, DC, RST, DELAY> InternalWiAdditions<SPI, CS, BUSY, DC, RST, DELAY>
+    for EPD7in5<SPI, CS, BUSY, DC, RST, DELAY>
 where
     SPI: Write<u8>,
     CS: OutputPin,
     BUSY: InputPin,
     DC: OutputPin,
     RST: OutputPin,
+    DELAY: DelayMs<u8>,
 {
-    fn init<DELAY: DelayMs<u8>>(
-        &mut self,
-        spi: &mut SPI,
-        delay: &mut DELAY,
-    ) -> Result<(), SPI::Error> {
+    fn init(&mut self, spi: &mut SPI, delay: &mut DELAY) -> Result<(), SPI::Error> {
         // Reset the device
         self.interface.reset(delay, 2);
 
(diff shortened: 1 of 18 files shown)


src/traits.rs
```diff
@@ -275,13 +280,14 @@ where
...
 ///```
-pub trait QuickRefresh<SPI, CS, BUSY, DC, RST>
+pub trait QuickRefresh<SPI, CS, BUSY, DC, RST, DELAY>
 where
     SPI: Write<u8>,
     CS: OutputPin,
     BUSY: InputPin,
     DC: OutputPin,
     RST: OutputPin,
+    DELAY: DelayMs<u8>,
```
caemor, line 290: Is the delay necessary for any of the QuickRefresh functions? And if it is we need to add the delay to these.

src/interface.rs
```diff
@@ -126,15 +130,32 @@ where
...
-            //old: shorten the time? it was 100 in the beginning
-            //self.delay_ms(5);
+            match task {
+                // TODO: Ignore this error?
+                Some((ref mut spi, ref mut delay, tcb)) => tcb(self, spi, delay)?,
+                None => {}
+            }
+            // //tested: REMOVAL of DELAY: it's only waiting for the signal anyway and should continue work asap
+            // //old: shorten the time? it was 100 in the beginning
+            // //self.delay_ms(5);
```
caemor, line 156: Since we need to insert a function here anyway wouldn't it be better to just "overwrite" the waveshare.wait_until_idle and handle everything there already (at a higher level) for these specific cases?

src/epd7in5_v2/mod.rs
```diff
@@ -223,8 +228,16 @@ where
...
+    fn wait_until_idle(&mut self, spi: &mut SPI, delay: &mut DELAY) -> Result<(), SPI::Error> {
+        self.interface.wait_until_idle(
+            IS_BUSY_LOW,
+            Some((spi, delay, |i, s, d| {
+                i.cmd(s, Command::GET_STATUS)?;
+                d.delay_ms(20);
+                Ok(())
+            })),
+        )?;
+        Ok(())
```
caemor, line 240: See the command in the interface. But wouldn't it be easier to just handle everything here already?

caemor (changes requested):
Thanks for the PR. 
The PR looks good. I'm a bit conflicted about the changes to the `Interface`- and `QuickRefresh`-Trait and would be happy to your feedback to my comments in these parts.

src/interface.rs
caemor: If not we should add a small doc example to highlight how it can be used (or link to the 7in5 v2 impl of yours above)

src/epd7in5_v2/mod.rs
olback: No idea why I made it this complicated.
Should be as easy as
```
while self.interface.is_busy(IS_BUSY_LOW) {
    self.interface.cmd(Command::GET_STATUS)?;
    delay.delay_ms(20);
}
Ok(())
```

caemor (approved):

caemor:
It would be great if you could reapply your changes/merge the main so the conflicts are resolved else I can do it on the weekend .

Thanks for your work!

PR #74, opened 2021-04-27
Fixed epd2in9 quick refresh

andber1:
I did some testing with my display and I noticed that the quick refresh does not work properly. There is ghosting. The remedy is to insert a hardware reset and to change the display routine (as in the original library). Now it works very fine.
However, I am not quite happy with the interface changes. I needed the delay parameter for the hardware reset and new function for displaying the new frame which is apparently not needed for the epd4in2 display. But I did not know how to solve this in a better way.
The interface for the quick refresh of the original waveshare library is:

- `void EPD_2IN9_V2_Display_Base(UBYTE *Image)`
  - This function calls `EPD_2IN9_V2_TurnOnDisplay` at the end and equals `update_old_frame` and `display_frame`
- `void EPD_2IN9_V2_Display_Partial(UBYTE *Image)`
  - This function calls `EPD_2IN9_V2_TurnOnDisplay_Partial` (with a display update control byte which is not documented in the specification) at the end and equals `update_new_frame` and `display_new_frame`
  - I do not know why they are calling this "partial". The whole display is updated. The real partial update is apparently not working.

### DIFF
--- a/src/epd2in9_v2/mod.rs
+++ b/src/epd2in9_v2/mod.rs
@@ -32,9 +32,21 @@
 //!    .into_styled(PrimitiveStyle::with_stroke(Black, 1))
 //!    .draw(&mut display);
 //!
-//!    // Display updated frame
+//!// Display updated frame
 //!epd.update_frame(&mut spi, &display.buffer(), &mut delay)?;
 //!epd.display_frame(&mut spi, &mut delay)?;
+//!
+//!// Draw something new here
+//!
+//!// Display new image as a base image for further quick refreshes
+//!epd.update_old_frame(&mut spi, &display.buffer(), &mut delay)?;
+//!epd.display_frame(&mut spi, &mut delay)?;
+//!
+//!// Update image here
+//!
+//!// quick refresh of updated pixels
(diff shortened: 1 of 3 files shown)


caemor (approved):

caemor:
They call it partial because its a partial refresh instead of a full refresh.
Partial Refresh is often much faster but there can be some slight ghosting after lots of uses, so you should use a full refresh every 10/100/1000 times. (best case you test it yourself after which amount of refreshes the contrast is worse enough to "waste" the time for a full refresh for your usecase).

This looks good to me and it looks like we are adding `delay` back to all the traits one after another (see #70 and #71) :smile: 
But there doesn't seem to be a way around it :shrug: 

Thanks for the PR. It looks good to me. If you don't think there is anything missing I'll merge it later today.

andber1:
I think this is complete now. It works very fine on my display. Thanks for your quick response and your approval.
I was confused by the "partial" update, because there is a function `update_partial_frame` in this crate which does something else.

caemor:
Yeah naming is difficult :see_no_evil: 
Thanks for your help!

PR #75, opened 2021-05-16
add support for 2in13bc tri-color e-ink display

lrbalt:
This PR adds support for Waveshare 2"13 tri-color (b/c) e-ink display. The datasheet is the same as for 2"9 b/c version. Added an example which uses the three colors

### DIFF
--- a/examples/epd2in13bc.rs
+++ b/examples/epd2in13bc.rs
@@ -23,6 +23,17 @@ use linux_embedded_hal::{
 // activate spi, gpio in raspi-config
 // needs to be run with sudo because of some sysfs_gpio permission problems and follow-up timing problems
 // see https://github.com/rust-embedded/rust-sysfs-gpio/issues/5 and follow-up issues
+//
+// This example first setups SPI communication using the pin layout found
+// at https://www.waveshare.com/wiki/2.13inch_e-Paper_HAT_(B). This example uses the layout for the
+// Raspberry Pi Zero (RPI Zero). The Chip Select (CS) was taken from the ep2in9 example since CE0 (GPIO8) did
+// not seem to work on RPI Zero with 2.13" HAT
+//
+// The first frame is filled with four texts at different rotations (black on white)
+// The second frame uses a buffer for black/white and a seperate buffer for chromatic/white (i.e. red or yellow)
+// This example draws a sample clock in black on white and two texts using white on red.
+//
+// after finishing, put the display to sleep
 
 fn main() -> Result<(), std::io::Error> {
     let busy = Pin::new(24); // GPIO 24, board J-18
@@ -104,7 +115,8 @@ fn main() -> Result<(), std::io::Error> {
         .into_styled(PrimitiveStyle::with_stroke(Black, 1))
         .draw(&mut display);
 
-    // draw white on Red background
+    // draw text white on Red background by using the chromatic buffer
+
     let _ = Text::new("It's working-WoB!", Point::new(90, 10))
         .into_styled(text_style!(
             font = Font6x8,
--- a/examples/epd2in13bc.rs
+++ b/examples/epd2in13bc.rs
@@ -1,0 +1,148 @@
+#![deny(warnings)]
+
+use embedded_graphics::{
+    fonts::{Font12x16, Font6x8, Text},
+    prelude::*,
+    primitives::{Circle, Line},
+    style::PrimitiveStyle,
+    text_style,
+};
+use embedded_hal::prelude::*;
+use epd_waveshare::{
+    color::*,
+    epd2in13bc::{Display2in13bc, Epd2in13bc},
+    graphics::{Display, DisplayRotation},
+    prelude::*,
+};
+use linux_embedded_hal::{
+    spidev::{self, SpidevOptions},
+    sysfs_gpio::Direction,
+    Delay, Pin, Spidev,
+};
+
+// activate spi, gpio in raspi-config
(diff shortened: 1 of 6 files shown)


examples/epd2in13bc.rs
```diff
@@ -0,0 +1,149 @@
...
+    graphics::{Display, DisplayRotation},
+    prelude::*,
+};
+use linux_embedded_hal::{
+    spidev::{self, SpidevOptions},
+    sysfs_gpio::Direction,
+    Delay, Pin, Spidev,
+};
+
+// activate spi, gpio in raspi-config
```
caemor, line 23: Please leave a few lines of what this display is showing either here, next to `main`, or at the top to make it easier for new people to see what this example is giving to them (e.g. color display example,...)

src/epd2in13bc/mod.rs
```diff
@@ -0,0 +1,389 @@
...
+//!    &mut spi,
+//!    &mono_display.buffer(),
+//!    &chromatic_display.buffer()
+//!)?;
+//!epd.display_frame(&mut spi, &mut delay)?;
+//!
+//!// Set the EPD to sleep
+//!epd.sleep(&mut spi, &mut delay)?;
+//!# Ok(())
+//!# }
```
caemor, line 54: Great to see the example here! :-)

src/epd2in13bc/mod.rs
```diff
@@ -0,0 +1,389 @@
...
+
+pub(crate) mod command;
+use self::command::Command;
+
+#[cfg(feature = "graphics")]
+mod graphics;
+#[cfg(feature = "graphics")]
+pub use self::graphics::Display2in13bc;
+
+/// Epd2in9bc driver
```
caemor, line 92:
```suggestion
/// Epd2in13bc driver
```

And maybe even a few more lines about its stats like width, height and other abilities specific for this one

src/epd2in13bc/mod.rs
```diff
@@ -0,0 +1,389 @@
...
+use crate::interface::DisplayInterface;
+use crate::traits::{
+    InternalWiAdditions, RefreshLut, WaveshareDisplay, WaveshareThreeColorDisplay,
+};
+
+/// Width of epd2in13bc in pixels
+pub const WIDTH: u32 = 104;
+/// Height of epd2in13bc in pixels
+pub const HEIGHT: u32 = 212;
+/// Default background color (white) of epd2in9bc display
```
caemor, line 70:
```suggestion
/// Default background color (white) of epd2in13bc display
```

caemor (changes requested):
Thanks for your PR :partying_face: 

Could you please add the display to the list in the readme as well?
And is everything working as expected when testing it?

examples/epd2in13bc.rs
lrbalt: I added some comments to the example

src/epd2in13bc/mod.rs
lrbalt: this was there already for epd2in9bc ;-)

src/epd2in13bc/mod.rs
lrbalt: oeps, fixed

lrbalt:
This driver was tested using a Raspberry Pi Zero with the display connected using the HAT for Raspberry Pi

lrbalt:
I've just rebased to main. @caemor do you think this can be merged?

caemor (approved):

caemor:
Looking good. Thanks for the PR!

lrbalt:
cool! Thanks

lrbalt:
I was thinking about support for TriColor. That way you can draw on a TriColor Display with black, white and chromatic, as defined in `src/color.rs`. Much easier than using two buffers.

looking at `/src/graphics.rs` this would mean adding `pub trait Display: DrawTarget<TriColor>`. This would be the third `Display` trait, besides `OctColor` and `BinaryColor`. 

The only challenge I see is the underlying buffer. To update the screen you need a b/w buffer for b/w and a b/w buffer for chromatic. The traits assume 1 buffer. Can I just add a `fn get_chromatic_buffer` to the `Display: DrawTarget<TriColor>` then?

Is this the way to go? What do you think @caemor ?

caemor:
Sounds good to me but don't you still need two buffers or will it just be a bigger single buffer instead?
And it will be an extension of https://github.com/caemor/epd-waveshare/blob/401858c4565343e9ec561cd9bfa61da3bc32c542/src/traits.rs#L51 for the display/graphics stuff?

I am looking forward to see the results of trying it out.

lrbalt:
There is the challenge I see. These traits currently ask for the buffer that the device understands. In this case, the trait assumes you need to update the screen twice (b/w and chromatic). 

If the display supports three colors, you'd like to update the TriColor Display and let the Device handle sending two buffers. The `update_color_frame` function needs both buffers as parameter, so this won't work.  

To start, I'm thinking of adding a function to the TriColor Display to return two buffers so you can call `update_color_frame` which does expose this device detail to the user. Perhaps I could use one buffer (twice the size) and just return two slices.

Perhaps there are users that like to handle the two buffers and call `update_*` for better performance, so removing the functions may not be the way to go.

We may consider adding a generic `fn update_frame_from_display(display)` to `WaveshareThreeColorDisplay` where display has a `fn get_buffer` and a `fn get_chromatic_buffer` or something. Of add it to the new `pub trait Display: DrawTarget<TriColor>`

Not sure. I'll experiment

PR #76, opened 2021-05-30
Enable drawing in three colors for epd2in13

lrbalt:
Move from BinaryColor to TriColor:  use one Display for drawing

* I use one buffer containing data for b/w and chromatic
* added two fn's to get slice for b/w buffer and slice for chromatic buffer for updating both buffers on the display
* you can still use just the b/w buffer if you want to use two colors
* updated the example to draw b/w and draw using three colors

### DIFF
--- a/src/epd2in13bc/graphics.rs
+++ b/src/epd2in13bc/graphics.rs
@@ -1,30 +1,33 @@
+use crate::color::TriColor;
 use crate::epd2in13bc::{DEFAULT_BACKGROUND_COLOR, HEIGHT, NUM_DISPLAY_BITS, WIDTH};
-use crate::graphics::{Display, DisplayRotation};
-use embedded_graphics::pixelcolor::BinaryColor;
+use crate::graphics::{DisplayRotation, TriDisplay};
 use embedded_graphics::prelude::*;
 
-/// Full size buffer for use with the 2in13b/c EPD
+/// Full size buffer for use with the 2.13" b/c EPD
 ///
 /// Can also be manually constructed and be used together with VarDisplay
 pub struct Display2in13bc {
-    buffer: [u8; NUM_DISPLAY_BITS as usize],
+    // one buffer for both b/w and for chromatic:
+    // * &buffer[0..NUM_DISPLAY_BITS] for b/w buffer and
+    // * &buffer[NUM_DISPLAY_BITS..2*NUM_DISPLAY_BITS] for chromatic buffer
+    buffer: [u8; 2 * NUM_DISPLAY_BITS as usize],
     rotation: DisplayRotation,
 }
 
 impl Default for Display2in13bc {
     fn default() -> Self {
         Display2in13bc {
-            buffer: [DEFAULT_BACKGROUND_COLOR.get_byte_value(); NUM_DISPLAY_BITS as usize],
+            buffer: [DEFAULT_BACKGROUND_COLOR.get_byte_value(); 2 * NUM_DISPLAY_BITS as usize],
             rotation: DisplayRotation::default(),
         }
     }
 }
 
-impl DrawTarget<BinaryColor> for Display2in13bc {
+impl DrawTarget<TriColor> for Display2in13bc {
     type Error = core::convert::Infallible;
 
-    fn draw_pixel(&mut self, pixel: Pixel<BinaryColor>) -> Result<(), Self::Error> {
-        self.draw_helper(WIDTH, HEIGHT, pixel)
+    fn draw_pixel(&mut self, pixel: Pixel<TriColor>) -> Result<(), Self::Error> {
+        self.draw_helper_tri(WIDTH, HEIGHT, pixel)
     }
 
     fn size(&self) -> Size {
@@ -32,7 +35,7 @@ impl DrawTarget<BinaryColor> for Display2in13bc {
     }
 }
 
-impl Display for Display2in13bc {
+impl TriDisplay for Display2in13bc {
     fn buffer(&self) -> &[u8] {
         &self.buffer
     }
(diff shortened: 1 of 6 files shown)


lrbalt:
I'd like to get your feedback @caemor before I update documentation

I'm wondering if you like to add a `fn` to `WaveshareThreeColorDisplay` which takes a buffer containing both b/w and chromatic data, i.e. `fn update_bw_and_chromatic_frame(&mut self, spi: &mut SPI, bw_and_chromatic_buffer: &[u8]`

lrbalt:
I've created a small application using both this patch and an emulator with one draw function. This way I can develop using emulator, knowing that it will work on the e-paper display too: https://github.com/lrbalt/epd-bootscreen

I needed to mock TriColor for using the emulator on my Mac (doesn't compile linux-embedded-hal), since it was defined in epd-waveshare, not embedded-graphics.

src/epd2in13bc/graphics.rs
```diff
@@ -1,38 +1,41 @@
...
-/// Full size buffer for use with the 2in13b/c EPD
+/// Full size buffer for use with the 2.13" b/c EPD
 ///
 /// Can also be manually constructed and be used together with VarDisplay
 pub struct Display2in13bc {
-    buffer: [u8; NUM_DISPLAY_BITS as usize],
+    // one buffer for both b/w and for chromatic:
+    // * &buffer[0..NUM_DISPLAY_BITS] for b/w buffer and
+    // * &buffer[NUM_DISPLAY_BITS..2*NUM_DISPLAY_BITS] for chromatic buffer
+    buffer: [u8; 2 * NUM_DISPLAY_BITS as usize],
```
caemor, line 13: I am not sure if this is really a good change, since it doubles the space requirements :thinking:

caemor (commented):
> I'm wondering if you like to add a fn to WaveshareThreeColorDisplay which takes a buffer containing both b/w and chromatic data, i.e. fn update_bw_and_chromatic_frame(&mut self, spi: &mut SPI, bw_and_chromatic_buffer: &[u8]
>
> and remove the individual update fn's for bw and chromatic from this trait. Or change the update_color_frrame to take one buffer. But that means that all tri-color screens need to use one big buffer for both b/w and chromatic. I.e. this exposes the internal data model in the api.

I don't think we should remove the option to just use a smaller buffer and update it in between sending b/w and chromatic. So allowing both options should be the best way to go forwards for now.

src/epd2in13bc/graphics.rs
lrbalt: To write to the epd, the data is sent using [interface::write](https://github.com/caemor/epd-waveshare/blob/401858c4565343e9ec561cd9bfa61da3bc32c542/src/interface.rs#L99) which takes a slice of the buffer. For tri-color a slice for the b/w-buffer or a slice to the white/chromatic buffer. This means that whatever datatype you choose here, I think you always need to convert it into those two buffers when updating the display probably allocating at each refresh

So I'm not sure how to avoid doubling space requirements.

lrbalt:
> > I'm wondering if you like to add a fn to WaveshareThreeColorDisplay which takes a buffer containing both b/w and chromatic data, i.e. fn update_bw_and_chromatic_frame(&mut self, spi: &mut SPI, bw_and_chromatic_buffer: &[u8]
> > and remove the individual update fn's for bw and chromatic from this trait. Or change the update_color_frrame to take one buffer. But that means that all tri-color screens need to use one big buffer for both b/w and chromatic. I.e. this exposes the internal data model in the api.
> 
> I don't think we should remove the option to just use a smaller buffer and update it in between sending b/w and chromatic. So allowing both options should be the best way to go forwards for now.

Ok, I'll add one new `fn` for updating both b/w and chromatic using the one buffer

src/epd2in13bc/graphics.rs
caemor: But aren't they written after each other? So there is still potentially the option to just replace the content of one of the buffers before sending it to the device. If that is not the case, we can just go forward with your change. And even if its not I think your change should still be fine, since a user with special needs should be able to use the VarDisplay instead :thinking:

lrbalt: True, but the buffer is not only used to drive the display but also to store the pixeldata when drawing before you send it to the display. You need a buffer that can hold tree colors when drawing. I mean, this display is for 3 colors?

If you want the user to handle two buffers themself, I don't see the point in adding `DrawTarget<TriColor>`. 

I'm not sure if we can define the buffer for only b/w in case of `DrawTarget<BinaryColor>` and twice that buffer in case of `DrawTarget<TriColor>`. Can we make `Display2in13bc` generic, i.e. `pub struct Display2in13bc<C: Color>` and use a C::BUFFER_SIZE constant or something?

caemor: > f you want the user to handle two buffers themself, I don't see the point in adding DrawTarget<TriColor>.
Good point!

> True, but the buffer is not only used to drive the display but also to store the pixeldata when drawing before you send it to the display. You need a buffer that can hold tree colors when drawing. I mean, this display is for 3 colors?

Yeah but you could also redraw the full screen. But I think you can just go forward with your change since it makes much more sense for the general use case.

lrbalt: ok, cool. I'll add the `fn` and fix up documentation.

lrbalt:
I've updated the example in `mod.rs` so all tests now pass. I did not add the `fn` mainly because it would look like:

```
fn update_bw_and_chromatic_frame(&mut self, spi: &mut SPI, bw_and_chromatic_buffer: &[u8]) -> Result<(), SPI::Error>;
```

This would mean that the trait `WaveshareThreeColorDisplay` would need to have knowledge of `Display2in13bc` to split the buffer. And we thus expose the inner data type to a trait. It doesn't feel right. It also led to conflicts on other displays that use some magic in the buffer for the chromatic part (like [here](https://github.com/caemor/epd-waveshare/blob/401858c4565343e9ec561cd9bfa61da3bc32c542/src/epd1in54b/mod.rs#L115)) so it would not really be generic.

caemor (approved):

lrbalt:
I think it is ready to merge.

lrbalt:
cool, thanks!

PR #81, opened 2021-07-14
Provide full QuickRefresh interface for 4.2 inch display

sirhcel:
The Waveshare 4.2 inch display needs no special commands for displaying the new frame. But it will come in handy to support the full `QuickRefresh` trait when it comes to supporting different quick refresh capable displays from an application.

This PR implements `display_new_frame` and `update_and_display_new_frame` using the already existing methods. It also brings some cleanup to warnings from `cargo doc` which showed up when checking the docs for the newly implemented methods.

### DIFF
--- a/src/epd4in2/mod.rs
+++ b/src/epd4in2/mod.rs
@@ -482,7 +482,7 @@ where
         Ok(())
     }
 
-    /// This is wrapper around `display_frame` for using this device as a true
+    /// This is a wrapper around `display_frame` for using this device as a true
     /// `QuickRefresh` device.
     fn display_new_frame(&mut self, spi: &mut SPI, delay: &mut DELAY) -> Result<(), SPI::Error> {
         self.display_frame(spi, delay)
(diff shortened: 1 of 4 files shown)


sirhcel:
And once again I forgot to say a big thank you for providing and maintaining this crate! :)

src/epd4in2/mod.rs
```diff
@@ -482,21 +482,24 @@ where
         Ok(())
     }
 
-    /// This function is not needed for this display
-    #[allow(unused)]
+    /// This is wrapper around `display_frame` for using this device as a true
```
caemor, line 485:
```suggestion
    /// This is a wrapper around `display_frame` for using this device as a true
```

caemor (changes requested):
Thanks for your PR.

caemor (approved):

caemor:
The improvements to the docs and the 4in2 display looks good to me so I will go forward and merge this already.

PR #83, opened 2021-08-13
Write data over SPI 1 byte at a time

### LINKED ISSUE #82: DisplayInterface::data should write one byte at a time
I've been trying to get the Waveshare 2.7" grayscale display working and I've been having some major issues, but I think the primary blocker is that DisplayInterface::data (and write) tries to write all the data at once rather than byte by byte, which breaks on some devices. 

See https://github.com/waveshare/e-Paper/blob/2a10334edcdd4e1c75eb74c7be3b6601614bef72/RaspberryPi_JetsonNano/python/lib/waveshare_epd/epd2in7.py#L164-L174 for the implementation on the 2.7". Note that the 2.7" chromatic display also uses the same code: https://github.com/waveshare/e-Paper/blob/2a10334edcdd4e1c75eb74c7be3b6601614bef72/RaspberryPi_JetsonNano/python/lib/waveshare_epd/epd2in7b.py#L111-L121.

There are some devices which seem to use a separate method to write a chunk of data at a time, but they seem to be the exception rather than the norm (https://github.com/waveshare/e-Paper/blob/2f1eeee9f5baa1753839cb1917f4e75b0c612620/RaspberryPi_JetsonNano/python/lib/waveshare_epd/epd7in5.py#L70-L74).

In my hacked up version, changing data to send bytes one at a time fixed the epd2in7. One of the issues I was seeing was that commands would hang when waiting for them to complete, in other cases it was displaying garbage, so it might fix #70, #68, and maybe others.

belak:
Fixes #82

This was the change I had to put in place to get the epd2in7 to start working. It's also closer in line with the provided drivers.

### DIFF
--- a/src/interface.rs
+++ b/src/interface.rs
@@ -57,10 +57,10 @@ where
     ///
     /// Enables direct interaction with the device with the help of [command()](Epd4in2::command())
     pub(crate) fn data(&mut self, spi: &mut SPI, data: &[u8]) -> Result<(), SPI::Error> {
-        for val in data.iter().copied() {
-            // high for data
-            let _ = self.dc.set_high();
+        // high for data
+        let _ = self.dc.set_high();
 
+        for val in data.iter().copied() {
             // Transfer data one u8 at a time over spi
             self.write(spi, &[val])?;
         }
@@ -90,9 +90,11 @@ where
         val: u8,
         repetitions: u32,
     ) -> Result<(), SPI::Error> {
+        // high for data
+        let _ = self.dc.set_high();
         // Transfer data (u8) over spi
         for _ in 0..repetitions {
-            self.data(spi, &[val])?;
+            self.write(spi, &[val])?;
         }
         Ok(())
     }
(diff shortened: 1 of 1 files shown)


src/interface.rs
```diff
@@ -57,11 +57,15 @@ where
     ///
     /// Enables direct interaction with the device with the help of [command()](Epd4in2::command())
     pub(crate) fn data(&mut self, spi: &mut SPI, data: &[u8]) -> Result<(), SPI::Error> {
-        // high for data
-        let _ = self.dc.set_high();
+        for val in data.iter().copied() {
+            // high for data
+            let _ = self.dc.set_high();
```
caemor, line 62: There shouldn't be a need to set dc high each time, since the output pin for dc stays on the same level if we don't change anything.

caemor:
Thanks for the PR. Going to look deeper into this later the upcoming weekend.

src/interface.rs
belak: Good call - I pushed a change for data and data_x_times which only sets dc high once.

caemor:
After I finally had time to check it and the difference was absolutely minimal (0.5% difference), where the measuring inaccuracy is most likely higher that the difference.

Thanks for your work and patience!

PR #85, opened 2021-08-20
Make embedded-graphics a dev dependency

jamwaffles:
e-g-core should be the only e-g dependency required for the driver itself, so this PR changes `Cargo.toml` and a few of the imports to reflect that.

Let me know if you want me to merge the imports and I'll do another pass to fix that up.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -16,19 +16,19 @@ edition = "2018"
 # travis-ci = { repository = "caemor/epd-waveshare" }
 
 [dependencies]
-embedded-graphics = { version = "0.7.1", optional = true}
 embedded-graphics-core = { version = "0.3.2", optional = true}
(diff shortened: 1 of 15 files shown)


caemor (approved):

caemor:
It looks good to me.
What do you mean with merging the imports?

jamwaffles:
> What do you mean with merging the imports?

As in, replacing

```rust
use foo::bar;
use foo::baz;
```

with

```rust
use foo::{bar, baz};
```

:)

caemor:
Sure go for it :+1:

jamwaffles:
So I thought I could get round to this but it turns out moving house takes more time and energy than I thought lol. Might be best to merge this PR without my cleanup to save everyone some time :)

caemor:
Thanks for the PR and good luck with moving! :smile:

PR #87, opened 2021-08-24
Add Color Conversions, Allow Changing Display Background Color

crzysdrs:
Adds some useful color conversions implementations so it's easier to work with the `image` and `embedded-graphics` crate.

Allow setting display background color (i.e. the border around the display), which I only recently realized was an option.

### DIFF
--- a/src/color.rs
+++ b/src/color.rs
@@ -1,9 +1,9 @@
 //! B/W Color for EPDs
 
 #[cfg(feature = "graphics")]
-use embedded_graphics::pixelcolor::BinaryColor;
+use embedded_graphics_core::pixelcolor::BinaryColor;
 #[cfg(feature = "graphics")]
-use embedded_graphics::pixelcolor::PixelColor;
+use embedded_graphics_core::pixelcolor::PixelColor;
 
 #[cfg(feature = "graphics")]
 pub use BinaryColor::Off as White;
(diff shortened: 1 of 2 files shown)


src/color.rs
```diff
@@ -72,9 +72,70 @@ impl From<()> for OctColor {
...
+    fn from(b: embedded_graphics::pixelcolor::raw::RawU4) -> Self {
+        use embedded_graphics::prelude::RawData;
+        OctColor::from_nibble(b.into_inner()).unwrap()
+    }
+}
+
 #[cfg(feature = "graphics")]
 impl PixelColor for OctColor {
-    type Raw = ();
+    type Raw = embedded_graphics::pixelcolor::raw::RawU4;
```
caemor, line 138: Did this change with the api or is this more efficient?

caemor:
Thanks for the PR!
Could you please switch to embedded_graphics_core (see #85)?
Everything else looks good and ready to merge to me.

src/color.rs
crzysdrs: It's actually just more correct, unrelated to API changes. It's basically stating that each `OctColor` pixel can be represented by 4 bits. This allows a raw buffer of 4 bit OctColor pixels to be readily converted into `ImageRaw`.
See [Raw Color Types Docs](https://docs.rs/embedded-graphics/0.7.1/embedded_graphics/pixelcolor/raw/index.html)

caemor: :+1:

caemor:
Is it ready to merge from your side?
From my side it looks good to go.

crzysdrs:
Yep, all good.

caemor:
Thanks for your work!

PR #95, opened 2022-01-17
DRAFT: Add support for epd2in13b_v3

Anshorei:
This still needs some cleaning up before merge, but the included example is confirmed to work.


caemor:
Do you think this is ready to integrate?

PR #73, opened 2021-04-18
Added epd2in9_v2 support

andber1:
I added support for the 2.9 display version 2. It is mostly copy paste from the version 1. I used the libraries provided by waveshare (https://www.waveshare.com/wiki/2.9inch_e-Paper_Module) to implement the changed behavior. Quick refresh is also supported and tested.

PR #103, opened 2022-03-30
epd2in9_v2 fixes

mike-kfed:
use LUT from waveshare C example code, speed up update_old_frame, copy
turn on display sequence from C-example

Hi,
I have a _Waveshare Pico-CapTouch-ePaper-2.9_ which has a 2.9" V2 display built in. I had some weird behaviour when updating the display (inverted colours, image flipped,...) so I investigated differences from the Rust- to their example C-code. With these changes my display works great. Since I am not sure what the idea behind this crates API is, I am of course open to adapt to your wishes.


src/epd2in9_v2/mod.rs
```diff
@@ -227,6 +227,7 @@ where
         Ok(())
     }
 
+    /// actually is the "Turn on Display" sequence
```
caemor, line 230: Is waveshare using a different sequence for the update?

caemor (approved):

src/epd2in9_v2/mod.rs
mike-kfed: Their code calls it like that. Also using this method makes the screen flicker, seems to be more than just updating, yes.

caemor:
Hi,
thanks for the update and improvements to the 2in9v2 display. It was introduced in #73 and was mostly a copy of the v1 display I believe. 
> Since I am not sure what the idea behind this crates API is, I am of course open to adapt to your wishes.

The idea is to be able to use different epd easily without having to do too much work to start using them. And its heavily relying on contributors who add and test the new display they own.

So I think your PR is ready to merge after the question above is resolved.

src/epd2in9_v2/mod.rs
caemor: Yeah the flicker is kind of normal to prevent ghosting (see https://electronics.stackexchange.com/questions/20276/why-does-flashing-prevent-ghosting-on-e-ink-displays for more infos)

mike-kfed: I have some little ghosting but I am fine with it doing this for fast updates:

```rust
        cur_epd
            .update_old_frame(&mut spi, empty_display.buffer(), &mut delay)
            .unwrap();
        cur_epd
            .update_new_frame(&mut spi, display.buffer(), &mut delay)
            .unwrap();
        cur_epd.display_new_frame(&mut spi, &mut delay).unwrap();
```

that mimicks also what their example C-code does in their "GUI" to update the icons. It however does the turn-on-sequence every 100th redraw or when the display fundamentally changes.

andber1:
@mike-kfed Thanks for your fixes. I must have missed this somehow. Your changes work fine on my display. Except one thing:
The change in line 382 leads to some artifacts on my display (some pixels are not updated). You have deleted the line `self.interface.cmd_with_data(spi, Command::WriteRam, buffer)?;` in the function `update_old_frame`. When I look in the original sources I cannot find that. That should correspond to the function `SetFrameMemory_Base`. See [here](https://github.com/waveshareteam/e-Paper/blob/af4d8b49ccef5f8f5fb88e9a836b86bd3f0bbfe3/Arduino/epd2in9_V2/epd2in9_V2.cpp#L425C11-L425C30).
Was this a mistake? If yes, I can create a new PR to fix that.

mike-kfed:
@andber1 feel free to open a PR. I was trying to mimic what the C code did back in the day, there's a good chance that I missed some detail. To this day I am happy with the drawing performance on my device, but that doesn't mean it works well for everyone.

PR #105, opened 2022-04-04
Fix width and height returned by size() when displays are rotated 90° and 270°

### LINKED ISSUE #104: When rotating displays dimensions are not getting updated
If you call `display.set_rotation(DisplayRotation::Rotate90);` or `display.set_rotation(DisplayRotation::Rotate270);` the display's dimensions (WIDTH/HEIGHT) should be swapped so that `embedded-graphics` text alignment features can work properly.  As things are right now, if you rotate the display 90° or 270° only left-aligned text will display properly.

riskable:
This PR fixes #104 where `embedded-graphics` text alignment features were not working because the `size()` method wasn't taking display rotation into account.

### DIFF
--- a/src/epd1in54/graphics.rs
+++ b/src/epd1in54/graphics.rs
@@ -39,7 +39,8 @@ impl DrawTarget for Display1in54 {
 
 impl OriginDimensions for Display1in54 {
     fn size(&self) -> Size {
-        Size::new(WIDTH, HEIGHT)
+        let (w, h) = self.dimensions();
+        Size::new(w.into(), h.into())
     }
 }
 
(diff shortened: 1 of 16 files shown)


src/epd1in54/graphics.rs
```diff
@@ -39,7 +39,8 @@ impl DrawTarget for Display1in54 {
 
 impl OriginDimensions for Display1in54 {
     fn size(&self) -> Size {
-        Size::new(WIDTH, HEIGHT)
+        let (w, h) = self.dimensions();
+        Size::new(w.into(), h.into())
```
caemor, line 43: The intos are not needed since w and h are already u32.

riskable:
@caemor Should I also change functions like `update_partial_frame()` to take the rotation into account so that the `x` and `y` variables get swapped if it's 270° or 90°?

caemor:
Sure that would be great!

caemor:
Should already be fixed.

Issue #102, opened 2022-03-21
1.54in has v2 version


buhe:
thanks..

caemor:
Closed by #106

Issue #79, opened 2021-06-13
Please help me with this almost-working example


lf94:
I'm trying to get Rust working on the Watchy: watchy.sqfmi.com .

I've gotten everything to compile and even flash, and even clear the screen! But nothing beyond that.

I'm wondering why drawing doesn't work at all. I'm not sure what to think - this is all new to me.

I've attached my code, in hopes that with some expertise, this is trivial to fix.

```
#![no_std]
#![no_main]

use xtensa_lx_rt as _;
use xtensa_lx_rt::entry;
use panic_halt as _;
use esp32_hal::{
  prelude::*,
  dport::Split,
  target,
  gpio::*,
  delay::Delay,
};

use epd_waveshare::{epd1in54::*, prelude::*};

#[entry]
fn main() -> ! {
  let p = target::Peripherals::take().unwrap();

  let (_, dport_clock_control) = p.DPORT.split();

  let clkcntrl = esp32_hal::clock_control::ClockControl::new(
      p.RTCCNTL,
      p.APB_CTRL,
      dport_clock_control,
      esp32_hal::clock_control::XTAL_FREQUENCY_AUTO,
  )
  .unwrap();

  let (clkcntrl_config, _) = clkcntrl.freeze().unwrap();

  let gpios = p.GPIO.split();
  let gpio18 = gpios.gpio18.into_push_pull_output();
  let gpio23 = gpios.gpio23.into_push_pull_output();
  let gpio5 = gpios.gpio5.into_push_pull_output();
  let gpio19 = gpios.gpio19.into_pull_up_input();
  let gpio10 = gpios.gpio10.into_push_pull_output();
  let gpio9 = gpios.gpio9.into_push_pull_output();

  let mut spi = esp32_hal::spi::SPI::<
    target::SPI3,
    Gpio18<Output<PushPull>>,
    Gpio23<Output<PushPull>>,
    Gpio1<Input<PullUp>>, // Not actually used...?
    Gpio5<Output<PushPull>>
  >::new(
    p.SPI3,
    esp32_hal::spi::Pins {
      sclk: gpio18,
      sdo: gpio23,
      sdi: None, // No SDI needed by the epaper display
      cs: None, // Used below?
    },
    esp32_hal::spi::config::Config {
      baudrate: 2_000_000.Hz(),
      data_mode: esp32_hal::spi::config::MODE_0,
      bit_order: esp32_hal::spi::config::BitOrder::MSBFirst,
    },
    clkcntrl_config,
[...]

PR #106, opened 2022-04-06
Add support for GDEH0154D67 (aka epd1in54_v2)

jcard0na:
The main changes from the previous version of the display, are the
lookup tables and the initialization procedures.  Also, partial vs. full
updates have a slightly different logic.

The changes were mostly ported from the vendor driver (EPAPER.c) as well
as Waveshare example driver (EPD_1in54_V2.cpp).  Both drivers are
released under public "use at will" license.

This was tested as:

```
use epd_waveshare::{epd1in54_v2::*, prelude::*};

let mut epd = Epd1in54::new(&mut spi, cs_pin, busy_in, dc, rst, &mut delay)?;

epd.set_lut(&mut spi, Some(RefreshLut::Full))?;
epd.clear_frame(&mut spi, &mut delay)?;
epd.display_frame(&mut spi, &mut delay)?;

// display cleared, yay!

let mut display = Display1in54::default();

let bmp: Bmp<BinaryColor> = Bmp::from_slice(light_data)?;
let image = Image::new(&bmp, Point::zero());
image.draw(&mut display)?;

epd.set_lut(&mut spi, Some(RefreshLut::Quick))?;
epd.update_frame(&mut spi, &display.buffer(), &mut delay)?;
epd.display_frame(&mut spi, &mut delay)?;

// display shows image
```

#102 #79


jcard0na:
One to the maintainer regarding this patch:

The file `graphics.rs` in this PR is an unmodified copy of the one in `epd1in54`.  I would be inclined to just link to that one instead, to avoid code duplication and ensure that future fixes and additions are applied to both displays.  What do you think?

caemor:
> Question to the maintainer regarding this patch:
> 
> The file `graphics.rs` in this PR is an unmodified copy of the one in `epd1in54`. I would be inclined to just link to that one instead, to avoid code duplication and ensure that future fixes and additions are applied to both displays. What do you think?

Yes this sounds good, if they behave the same way.

src/type_a/constants.rs
```diff
@@ -16,6 +16,53 @@ pub(crate) const LUT_PARTIAL_UPDATE: [u8; 30] =[
...
+    0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 
+    0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 
+    0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 
+    0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 
+    0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 
+    0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 
+    0x22, 0x22, 0x22, 0x22, 0x22, 0x22, 0x0, 0x0, 0x0, 
+    0x02, 0x17, 0x41, 0xB0, 0x32, 0x28, 
+];
+
```
caemor, line 65: I think it would be better to move the constants next to the epd1in54 v2 since they seem to be specific for this display (at least I didn't found the same sequence in here)

src/lib.rs
```diff
@@ -76,6 +76,7 @@ mod interface;
 pub mod epd1in54;
 pub mod epd1in54b;
 pub mod epd1in54c;
+pub mod epd1in54_v2;
```
caemor, line 79: A `cargo fmt` might be necessary here

src/epd1in54_v2/graphics.rs
```diff
@@ -0,0 +1,162 @@
+use crate::epd1in54::{DEFAULT_BACKGROUND_COLOR, HEIGHT, WIDTH};
```
caemor, line 1: Just use the v1 one if there is no difference.

caemor (changes requested):
Thanks for the PR. 
It nearly looks ready to merge besides the move of the graphics file and the LUTs.

jcard0na:
I believe I addressed all the change requests.  Let me know if there's anything else that should be addressed before merging.

caemor:
I think the new constants/LUTs file might be missing

jcard0na:
Duh!  Sorry about that.  Just added it.

caemor (approved):

lf94:
:o!!! Awesome work! Thank you!

PR #108, opened 2022-04-18
Remove `background_color` and `clear_frame`

### LINKED ISSUE #100: `background_color` is misused on color displays
For some reason, all of the color displays seem to be using their background color as the color to set the chromatic layer to when rendering in black-and-white. This means that if their background color is set to black, the whole display gets set to red. (Also, after doing that by accident, it seems to have permanently washed out my display's red color; it might be because it was trying to set a pixel to black and red at the same time?)
On top of that, a lot of them don't actually use it as the color to clear the display in `clear_frame` either, just using the default background color.

As an aside, `background_color` seems a bit odd in general; it's only used for `clear_frame`, and doesn't actually affect the background color when rendering with `embedded-graphics`. Is it really needed?

Liamolucko:
Resolves #100 

Removes `background_color`, `set_background_color`, and `clear_frame`, and properly handles the filling of the chromatic layer when rendering in black and white.

Looking back at #100, I'm not entirely clear if you wanted to keep or remove `clear_frame`; I'm happy to put it back taking the color to clear with as an argument.

I did find one display which was using `background_color` for stuff other than `clear_frame` - the epd5in16f was using it to set the border color. For now, I've replaced it with a `set_border_color` method which is only on that display. It should eventually be replaced by a proper method on `WaveshareDisplay`, but I don't have any hardware which supports border colors that I can test with.


caemor:
Sorry for the late response I just realized that I never had published my review and it was still pending :see_no_evil:

peckpeck:
I think we should just keep the clear_frame method, it is useful for initializing and testing the epd, especially when compiled without the graphic feature

PR #118, opened 2022-10-14
Add epd7in5b_v2 support

peckpeck:
Support of 7.5" (b) v2 (3 colors with back, white, red)
It is based on  epd7in5_v2 which doesn't work (epd7in5_v3 neither)
It is also based on the arduino C driver and datasheet.
It works on my V3 but the C driver and the datasheet haven't changed between v2 and v3.

### DIFF
--- a/src/epd7in5b_v2/command.rs
+++ b/src/epd7in5b_v2/command.rs
@@ -1,0 +1,156 @@
+//! SPI Commands for the Waveshare 7.5"(B) V2 E-Ink Display
+
+use crate::traits;
+
+/// Epd7in5 commands
+///
+/// Should rarely (never?) be needed directly.
+///
+/// For more infos about the addresses and what they are doing look into the PDFs.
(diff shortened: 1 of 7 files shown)


src/epd7in5b_v2/command.rs
```diff
@@ -0,0 +1,156 @@
+//! SPI Commands for the Waveshare 7.5"(B) V2 E-Ink Display
```
caemor, line 1:
```suggestion
//! SPI Commands for the Waveshare 7.5"(B) V2 and V3 E-Ink Display
```

Maybe include the v3 as well then if its already tested and working with the same commands.

caemor (changes requested):
Could you please add the display(s) to the changelog and the readme as well?

peckpeck:
Done

caemor (approved):
Thanks

PR #125, opened 2022-11-07
Make wait_for_idle work for everyone, with busy loop or sleep loop

peckpeck:
The delay structure is missing in many places where wait_until_idle is needed, so i added it everywhere.

Moreover, the default is to busy loop, but it can be very interesting on microcontrolers to sleep instead, either to save energy or to give time to another task. Especially with large displays where the refresh time can take several seconds. So I made the sleep time a parameter the user can control.

I also made `wait_until_idle` public, because, even if methods call it when it is mandatory, the user can still want to call it to make sure last data has been processed before going on.

### DIFF
--- a/src/epd5in65f/mod.rs
+++ b/src/epd5in65f/mod.rs
@@ -96,8 +96,9 @@ where
         dc: DC,
         rst: RST,
         delay: &mut DELAY,
+        delay_ms: u8,
     ) -> Result<Self, SPI::Error> {
-        let interface = DisplayInterface::new(cs, busy, dc, rst);
+        let interface = DisplayInterface::new(cs, busy, dc, rst, delay_ms);
         let color = DEFAULT_BACKGROUND_COLOR;
 
         let mut epd = Epd5in65f { interface, color };
@@ -120,9 +121,9 @@ where
         &mut self,
         spi: &mut SPI,
         buffer: &[u8],
-        _delay: &mut DELAY,
+        delay: &mut DELAY,
     ) -> Result<(), SPI::Error> {
-        self.wait_busy_high();
+        self.wait_until_idle(spi, delay)?;
         self.update_vcom(spi)?;
         self.send_resolution(spi)?;
         self.cmd_with_data(spi, Command::DataStartTransmission1, buffer)?;
@@ -132,6 +133,7 @@ where
     fn update_partial_frame(
         &mut self,
         _spi: &mut SPI,
+        _delay: &mut DELAY,
         _buffer: &[u8],
         _x: u32,
         _y: u32,
@@ -141,14 +143,14 @@ where
         unimplemented!();
     }
 
-    fn display_frame(&mut self, spi: &mut SPI, _delay: &mut DELAY) -> Result<(), SPI::Error> {
-        self.wait_busy_high();
+    fn display_frame(&mut self, spi: &mut SPI, delay: &mut DELAY) -> Result<(), SPI::Error> {
+        self.wait_until_idle(spi, delay)?;
         self.command(spi, Command::PowerOn)?;
-        self.wait_busy_high();
+        self.wait_until_idle(spi, delay)?;
         self.command(spi, Command::DisplayRefresh)?;
-        self.wait_busy_high();
+        self.wait_until_idle(spi, delay)?;
         self.command(spi, Command::PowerOff)?;
-        self.wait_busy_low();
+        self.wait_busy_low(delay);
         Ok(())
     }
 
@@ -165,7 +167,7 @@ where
 
     fn clear_frame(&mut self, spi: &mut SPI, delay: &mut DELAY) -> Result<(), SPI::Error> {
         let bg = OctColor::colors_byte(self.color, self.color);
-        self.wait_busy_high();
+        self.wait_until_idle(spi, delay)?;
         self.update_vcom(spi)?;
         self.send_resolution(spi)?;
         self.command(spi, Command::DataStartTransmission1)?;
@@ -193,13 +195,15 @@ where
     fn set_lut(
         &mut self,
         _spi: &mut SPI,
+        _delay: &mut DELAY,
         _refresh_rate: Option<RefreshLut>,
     ) -> Result<(), SPI::Error> {
         unimplemented!();
     }
 
-    fn is_busy(&self) -> bool {
-        self.interface.is_busy(IS_BUSY_LOW)
+    fn wait_until_idle(&mut self, _spi: &mut SPI, delay: &mut DELAY) -> Result<(), SPI::Error> {
+        self.interface.wait_until_idle(delay, IS_BUSY_LOW);
+        Ok(())
     }
 }
 
@@ -229,11 +233,10 @@ where
         self.interface.cmd_with_data(spi, command, data)
     }
 
-    fn wait_busy_high(&mut self) {
-        self.interface.wait_until_idle(true);
-    }
-    fn wait_busy_low(&mut self) {
-        self.interface.wait_until_idle(false);
+    fn wait_busy_low(&mut self, delay: &mut DELAY) {
+        while self.interface.is_busy(false) {
+            delay.delay_ms(5);
+        }
     }
     fn send_resolution(&mut self, spi: &mut SPI) -> Result<(), SPI::Error> {
         let w = self.width();
--- a/src/epd5in65f/mod.rs
+++ b/src/epd5in65f/mod.rs
@@ -96,7 +96,7 @@ where
         dc: DC,
         rst: RST,
         delay: &mut DELAY,
-        delay_ms: u8,
+        delay_ms: Option<u8>,
     ) -> Result<Self, SPI::Error> {
         let interface = DisplayInterface::new(cs, busy, dc, rst, delay_ms);
         let color = DEFAULT_BACKGROUND_COLOR;
--- a/src/epd5in65f/mod.rs
+++ b/src/epd5in65f/mod.rs
@@ -234,9 +234,7 @@ where
     }
 
     fn wait_busy_low(&mut self, delay: &mut DELAY) {
-        while self.interface.is_busy(false) {
-            delay.delay_ms(5);
-        }
+        self.interface.wait_until_idle(delay, false);
     }
     fn send_resolution(&mut self, spi: &mut SPI) -> Result<(), SPI::Error> {
         let w = self.width();
--- a/src/epd7in5_v2/mod.rs
+++ b/src/epd7in5_v2/mod.rs
@@ -96,8 +96,9 @@ where
         dc: DC,
         rst: RST,
         delay: &mut DELAY,
+        delay_ms: u8,
     ) -> Result<Self, SPI::Error> {
-        let interface = DisplayInterface::new(cs, busy, dc, rst);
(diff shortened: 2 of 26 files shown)


mangelajo:
+1 to this change.

I am having issues under the esp32 platform with std, running the display inside a thread I get errors from the underlying FreeRTOS because of the busy loop:
```
I (317323) esp-x509-crt-bundle: Certificate validated
Display: 
Display: 31  PLAZA MAYOR     1m 29s
Display: 31  PLAZA MAYOR     1m 29s
Display: 39  PLAZA ESPAÑA    3m 42s
Display: 65  BENAVENTE       4m 50s
Display: 36  ATOCHA          10m 41s
Display: 39  PLAZA ESPAÑA    12m 08s
Display: 65  BENAVENTE       13m 41s
Display: 33  PRINCIPE PIO    17m 34s
Display: 36  ATOCHA          19m 53s
Display: 33  PRINCIPE PIO          
Display: *
E (325323) task_wdt: Task watchdog got triggered. The following tasks did not reset the watchdog in time:
E (325323) task_wdt:  - IDLE (CPU 0)
E (325323) task_wdt: Tasks currently running:
E (325323) task_wdt: CPU 0: pthread
E (325323) task_wdt: CPU 1: IDLE
E (325323) task_wdt: Print CPU 0 (current core) backtrace


Backtrace: 0x42098712:0x3FC989A0 0x4037982D:0x3FC989C0 0x42015145:0x3FCEE0B0 0x4201516E:0x3FCEE0D0 0x42007B60:0x3FCEE0F0 0x4200AB32:0x3FCEE120 0x4200AC56:0x3FCEE140 0x4200749B:0x3FCEE160 0x4200863B:0x3FCEE250 0x420085FF:0x3FCEE290 0x4200B1AF:0x3FCEE2D0 0x4200A294:0x3FCEE310 0x4206127F:0x3FCEE350 0x420612A6:0x3FCEE370 0x4205A050:0x3FCEE390 0x42077B04:0x3FCEE3B0

E (325323) task_wdt: Print CPU 1 backtrace
```

I was looking into fixing it, but I see you handled it already, thanks.

src/interface.rs
```diff
@@ -134,13 +137,17 @@ where
...
+    pub(crate) fn wait_until_idle(&mut self, delay: &mut DELAY, is_busy_low: bool) {
         while self.is_busy(is_busy_low) {
-            // //tested: REMOVAL of DELAY: it's only waiting for the signal anyway and should continue work asap
-            // //old: shorten the time? it was 100 in the beginning
-            // //self.delay_ms(5);
+            // This has been removed and added many time :
+            // - it is faster to not have it
+            // - it is complicated to pass the delay everywhere all the time
+            // - busy waiting can consume more power that delaying
+            // - delay waiting enables task switching on realtime OS
```
mangelajo, line 146: exactly what I need!, thanks

peckpeck: Be careful, on esp32 you have to use delay::FreeRtos and not delay::Ets and a minimum delay_ms on 10, to make sure there is a task switch

mangelajo: ah, thanks, I knew and tried delay::FreeRtos, but didn't know about the 10ms limitation.

src/traits.rs
```diff
@@ -147,6 +160,7 @@ where
         dc: DC,
         rst: RST,
         delay: &mut DELAY,
+        delay_ms: u8,
```
caemor, line 163: 1. Would it make sense to make this an option? (longer explanation in the final comment)
```suggestion
        delay_ms: Option<u8>,
```

src/interface.rs
```diff
@@ -31,14 +33,15 @@ where
     RST: OutputPin,
     DELAY: DelayMs<u8>,
 {
-    pub fn new(cs: CS, busy: BUSY, dc: DC, rst: RST) -> Self {
+    pub fn new(cs: CS, busy: BUSY, dc: DC, rst: RST, delay_ms: u8) -> Self {
```
caemor, line 36: 2. This could then be either read from a default value/constant for this specific epd or from the optional delay_ms in WaveshareDisplay::new()

caemor (changes requested):
I like the new change, but currently a few displays define their own wait_until_idle without using the given delay_ms. It would great to use it there as well and the case `delay_ms` is None fall back to a sensible constant (the current delay)

src/epd7in5b_v2/mod.rs
```diff
@@ -221,13 +223,24 @@ where
...
-        self.interface.is_busy(IS_BUSY_LOW)
+    /// wait
+    fn wait_until_idle(&mut self, spi: &mut SPI, delay: &mut DELAY) -> Result<(), SPI::Error> {
+        // C driver first sends the command
+        self.interface.cmd(spi, Command::GetStatus)?;
+        while self.interface.is_busy(IS_BUSY_LOW) {
+            self.interface.cmd(spi, Command::GetStatus)?;
+            // C driver doesn't wait here but we have to
+            // because be want to be able to give back control if we are in a RT thread
+            delay.delay_ms(10);
```
caemor, line 240: 3. Which could be used here as well
```suggestion
            delay.delay_ms(self.delay_ms);
```

peckpeck: If you look at the comment, the goal is to give back control to scheduler in case of a RT thread, this means the value needed here depends on the host OS, so it should be under the control of the library user, not the developer.
So, I indeed think that those calls to should use `self.delay_ms` but I also think most customized `wait_until_idle` should disappear in favor of a common one in interface. There are only 2 flavors, one with a call to GetStatus and one without.

peckpeck: Anyway, having a default value for users who don't care is also interesting

src/traits.rs
```diff
@@ -139,6 +149,10 @@ where
     type DisplayColor;
     /// Creates a new driver from a SPI peripheral, CS Pin, Busy InputPin, DC
     ///
+    /// delay_ms is the number of ms the idle loop should sleep on.
```
caemor, line 152:
```suggestion
    /// `delay_ms` is the number of ms the idle loop should sleep on.
```

src/interface.rs
```diff
@@ -31,14 +33,17 @@ where
     RST: OutputPin,
     DELAY: DelayMs<u8>,
 {
-    pub fn new(cs: CS, busy: BUSY, dc: DC, rst: RST) -> Self {
+    pub fn new(cs: CS, busy: BUSY, dc: DC, rst: RST, delay_ms: Option<u8>) -> Self {
```
caemor, line 36:
```suggestion
    /// Creates a new DisplayInterface struct
    /// 
    /// If no delay is given a default delay of 10ms is used.
    pub fn new(cs: CS, busy: BUSY, dc: DC, rst: RST, delay_ms: Option<u8>) -> Self {
```

src/epd7in5b_v2/mod.rs
```diff
@@ -119,8 +119,9 @@ where
         dc: DC,
         rst: RST,
         delay: &mut DELAY,
+        delay_ms: Option<u8>,
     ) -> Result<Self, SPI::Error> {
-        let interface = DisplayInterface::new(cs, busy, dc, rst);
+        let interface = DisplayInterface::new(cs, busy, dc, rst, delay_ms);
```
caemor, line 124: Since the default delay is the same delay as the one used in here (10ms) nothings needs to be changed

src/epd7in5_v3/mod.rs
```diff
@@ -151,8 +158,9 @@ where
         dc: DC,
         rst: RST,
         delay: &mut DELAY,
+        delay_ms: Option<u8>,
     ) -> Result<Self, SPI::Error> {
-        let interface = DisplayInterface::new(cs, busy, dc, rst);
+        let interface = DisplayInterface::new(cs, busy, dc, rst, delay_ms);
```
caemor, line 163: Optional:
```suggestion
        let interface = DisplayInterface::new(cs, busy, dc, rst, delay_ms.unwrap_or(DEFAULT_DELAY);
```
Where Default Delay is the old 20ms delay from below.

src/epd7in5_v2/mod.rs
```diff
@@ -96,8 +96,9 @@ where
         dc: DC,
         rst: RST,
         delay: &mut DELAY,
+        delay_ms: Option<u8>,
     ) -> Result<Self, SPI::Error> {
-        let interface = DisplayInterface::new(cs, busy, dc, rst);
+        let interface = DisplayInterface::new(cs, busy, dc, rst, delay_ms);
```
caemor, line 101: Same optional change as epd7in5_v3

src/epd5in65f/mod.rs
```diff
@@ -229,11 +233,10 @@ where
...
     }
 
-    fn wait_busy_high(&mut self) {
-        self.interface.wait_until_idle(true);
-    }
-    fn wait_busy_low(&mut self) {
-        self.interface.wait_until_idle(false);
+    fn wait_busy_low(&mut self, delay: &mut DELAY) {
+        while self.interface.is_busy(false) {
+            delay.delay_ms(5);
```
caemor, line 238: Whats the reasoning for the 5ms here instead of the 10ms we used elsewhere?

caemor (changes requested):
Thanks for the great work. 
Do you think it would make sense to include the DelayUS update in here as well?

src/epd7in5b_v2/mod.rs
peckpeck: The parameter must still be passed

src/epd7in5_v3/mod.rs
peckpeck: the default value is handled inside interface

src/epd5in65f/mod.rs
peckpeck: The reasoning it that I forgot this case :-)

peckpeck:
Fixing DelayUs touches a lot of files, I think it would be more readable in a separate PR.

PR updated

src/epd5in65f/mod.rs
```diff
@@ -229,11 +233,8 @@ where
         self.interface.cmd_with_data(spi, command, data)
     }
 
-    fn wait_busy_high(&mut self) {
-        self.interface.wait_until_idle(true);
-    }
-    fn wait_busy_low(&mut self) {
-        self.interface.wait_until_idle(false);
+    fn wait_busy_low(&mut self, delay: &mut DELAY) {
+        self.interface.wait_until_idle(delay, false);
```
caemor, line 237:
```suggestion
        self.interface.wait_until_idle(delay, !IS_BUSY_LOW);
```
Alternatively we could remove the IS_BUSY_LOW constant here and just use true and false for `ẁait_until_idle` and `wait_busy_low` to be a bit more consistent.

src/epd7in5_v3/mod.rs
caemor: I was referring the old default for this specific display which was 20ms before

peckpeck: I don't think there is any value in this default. The delay tells how long to wait before looping, and this is linked to cpu usage and not to the device.

src/epd5in65f/mod.rs
caemor: Since its not that important, I'm just gonna merge this now.

caemor (approved):

PR #123, opened 2022-11-03
Make Display more generic and remove lot of redundant code

peckpeck:
As suggested in #121

PR #126, opened 2022-11-19
Add driver for epd2in13_v3 based on v2

fmeef:
It seems the only difference between the V2 and V3 drivers is a change in the LUT data. This is sadly undocumented in the official reference implementation.


src/epd2in13_v3/command.rs
```diff
@@ -0,0 +1,276 @@
...
+        }
+    }
+}
+
+impl traits::Command for Command {
+    /// Returns the address of the command
+    fn address(self) -> u8 {
+        self as u8
+    }
+}
```
caemor, line 276: I haven't checked it yet, but if v2 and v3 only differ in the lut constants we could fully reuse this here instead

src/epd2in13_v3/graphics.rs
```diff
@@ -0,0 +1,171 @@
...
+
+        let buffer = display.buffer();
+
+        assert_eq!(buffer[0], Color::Black.get_byte_value());
+
+        for &byte in buffer.iter().skip(1) {
+            assert_eq!(byte, epd2in13_v2::DEFAULT_BACKGROUND_COLOR.get_byte_value());
+        }
+    }
+}
```
caemor, line 171: With @peckpeck pr #123  merged this isn't necessary anymore

src/epd2in13_v3/mod.rs
```diff
@@ -0,0 +1,577 @@
...
+        HEIGHT
+    }
+
+    fn set_lut(
+        &mut self,
+        spi: &mut SPI,
+        refresh_rate: Option<RefreshLut>,
+    ) -> Result<(), SPI::Error> {
+        let buffer = match refresh_rate {
+            Some(RefreshLut::Full) | None => &LUT_FULL_UPDATE,
```
caemor, line 352: Optional: We could merge v2 and v3 if we use a feature gate or sth similar, which would reduce the duplicated code by a lot.

caemor (changes requested):

caemor:
Thanks for your PR :partying_face: 

There are a few things, where we could reduce code duplication I believe.

src/epd2in13_v3/mod.rs
peckpeck: I would suggest merging everything into the same structure and just having a different `new` constructor.

If you insist on having 2 different structures, you can add a generic parameter or a const generic to make them different types.

fmeef:
The new generic graphics::Display seems to be having issues not present in the old implementation I was using before rebasing on main. Running update_and_display_frame() (or anything that calls update_frame()) results in a assertion failure

```
thread 'main' panicked at 'assertion failed: buffer.len() == buffer_len(WIDTH as usize, HEIGHT as usize)', /workspace/epd-waveshare/src/epd2in13_v3/mod.rs:210:9
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
abort() was called at PC 0x4202161b on core 0
```

what is interesting here is the assertion's buffer_length is calculated with
```rust
(width +7) / 8 * height
```

which can't be true in the case of the V2 code, as the buffer length in v2 is 
```rust
{ WIDTH as usize * HEIGHT as usize / 8 }
```

peckpeck:
@fmeef that is interesting since the documentation on other epd says the buffer is aligned on lines, and this epd specifically has a number of pixels per line that is not a multiple of 8.

So the question is : do you know if your display splits the last byte of the first line? ie do some bits of the 16th byte of your buffer display pixels on the second line ?

peckpeck:
After looking at the datasheet, it looks like the the driver has swapped the definition of width and height.
Moreover the buffer_len function is only used there. I think it should be removed.

src/epd2in13_v3/mod.rs
caemor: Different constructor sounds like a good idea

peckpeck:
Since it worked before, and since this byte rounding was done in `Display` as well as in buffer_len, we can answer the previous question, the byte is not split, some bits are just unused.

So we now have 2 possibilities : 
* swap width and height in this driver (it is the only one having this property and it supports the change at hardware level)
* make use of `buffer_len` everywhere and fix the generic `Display` to match

The first one use less code but needs a device for testing, the second one is more generic.

caemor:
```Rust
/// Width of the display.
pub const WIDTH: u32 = 122;

/// Height of the display
pub const HEIGHT: u32 = 250;
```
As long as the line isn't split we can't use WIDTH / 8 * HEIGHT (=3750 Bytes) (or even WIDTH*HEIGHT / 8 = 3812.5 Bytes) since this rounds down and then we would be missing a few bits and maybe even bytes.
But we need 4000 Bytes to roundup to the full linewidth:
grafik

Which equals `(width +7) / 8 * height`:
grafik

peckpeck:
o yes, i thought 250/8 was an integer, silly me.
So I must fix the generic display, and we should use `buffer_len` everywhere, even when not necessary to avoid forgetting it.

peckpeck:
Fix: https://github.com/caemor/epd-waveshare/pull/128

fmeef:
Thanks, I updated the PR to reflect those changes. Now it gets past the assert fine, but after refreshing the display it is entirely black (which didn't happen with the old graphics module). I am debugging this and will get back in a bit when I have an idea what happened. Changing the BWRBIT parameter didn't help, but I didn't expect it to since this isn't a tricolor display.

peckpeck:
2 things can have this effect: 
* now a clear_frame may be needed after a new since the `Display::new` color is the same for all devices, this may have changed after the last PR
* the color type has also been unified, it is now a `Color` and not a `BinaryColor` maybe the meaning of black and white have been mixed up on some devices

fmeef:
Figured it out, there was a slight change in behavior in #128 that my testing code didn't account for. Needed to explicitly call .clear() with the background color for the buffer to be initialized correctly.

caemor:
Thanks for your work, @fmeef!
Does this work for you now?
And could you maybe test #138 as an alternative approach?
You would just need to change adapt the cargo.toml to sth like:
```
epd-waveshare = { git="https://github.com/caemor/epd-waveshare", branch= "2in13v3-featuregate", default-features = false, features = ["graphics", "linux-dev", "epd2in13_v3"] }
```

caemor:
Thanks for your work. This got succeeded by #138.

PR #134, opened 2022-12-05
Remove background color but not clear_frame

peckpeck:
Sorry for @Liamolucko 
I could not start from https://github.com/caemor/epd-waveshare/pull/108 it was too much work to re-add clear_frame

### DIFF
--- a/src/epd7in5_v3/mod.rs
+++ b/src/epd7in5_v3/mod.rs
@@ -2,9 +2,9 @@
 //!
 //! # References
 //!
-//! - [Datasheet](https://www.waveshare.com/wiki/7.5inch_e-Paper_HAT)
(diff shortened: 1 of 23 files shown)


src/epd7in5_v3/mod.rs
```diff
@@ -23,6 +23,8 @@ use self::command::Command;
 use crate::buffer_len;
 
 /// Full size buffer for use with the 7in5 v3 EPD
+// Note v2 is bicolor, cannot find v3 on the website
+// 7in5B v3 exists and is tricolor, but it doesn't work wit this driver
```
caemor, line 27: I believe it was b v3 (https://www.waveshare.com/wiki/7.5inch_e-Paper_HAT_(B)_Manual#Introduction) (https://github.com/caemor/epd-waveshare/issues/26#issuecomment-1132293530). It was based of marcelbuesings original work for the 7in5bc  and I though he got it working. Did you test it?
The module should be renamed to `epd7in5b_v3` anyway

PR #135, opened 2022-12-06
Add conversion from BinaryColor to Color

jcard0na:
Without this we get an error when drawing BinaryColor images onto epd1in54_v2 and probably other BW displays.


jcard0na:
Oh, I see that commit b852f2d31a126de39e1c95f14f4ebd79be7847fd added this change already, so will drop this PR.

PR #139, opened 2022-12-08
[epd1in54_v2] Do enter deep sleep when epd.sleep() is invoked

jcard0na:
The documentation for [`WaveshareDisplay::sleep()`](https://docs.rs/epd-waveshare/latest/epd_waveshare/prelude/trait.WaveshareDisplay.html#tymethod.sleep) states that the function puts the device in deep sleep.  According to the documentation for this display, this requires sending a 0x1.  Tested on a device, and observed a drop of standby current from ~10 uA to ~1 uA.

This probably applies to all the displays that internally use the SSD1681 controller, but I have no way to test.


caemor (approved):

caemor:
Thanks, that sounds like a great improvement.

PR #141, opened 2023-01-07
WIP: Fix 7in5(HD) by allowing blockwise data to be written

### LINKED ISSUE #142: Bytewise vs Blockwise writing of data over SPI
Got a chance recently to take a look at this and I believe the culprit might be the change made in #83 which writes bytes one at time instead of all at once as reverting this got my 7.5" HD display working again and suspect this may be the case for the V2 too.
        Happy to make a PR to sort this one out but seems like we may need to support both depending on the display, don't suppose we know which screens will need each write method?

_Originally posted by @whiite in https://github.com/caemor/epd-waveshare/issues/70#issuecomment-1316680214_

caemor:
This pr adds a possible fix for a bug introduced  in #83 and discussed in #70.
A switch case is added to let each epd decide if it wants to write its data blockwise or bytewise over spi.

I wasn't able to test this myself.

Closes #142.

### DIFF
--- a/src/epd1in54/mod.rs
+++ b/src/epd1in54/mod.rs
@@ -52,6 +52,7 @@ pub const HEIGHT: u32 = 200;
 pub const DEFAULT_BACKGROUND_COLOR: Color = Color::White;
 //const DPI: u16 = 184;
 const IS_BUSY_LOW: bool = false;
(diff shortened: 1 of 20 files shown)


caemor:
I'm going to merge this now since this hopefully resolves the issue although it couldn't be tested yet.

PR #144, opened 2023-01-26
Add EPD 4in2bc support

yurinnick:
So... Does it work? Well, mostly. Display works fine as far as I can tell, however color coding is broken. For some reason `4in2bc` uses a completely different scheme than anything implemented.

```
// 4in2bc
// White bw - 0xFF c - 0xFF
// Black bw - 0 c - 0xFF
// Red bw - 0 c - 0

// bwrbit = true
// White: bw - 0xFF c - 0
// Black: bw - 0 c - 0
// Red: bw - 0 c - 0xFF

// bwrbit = false
// White: bw - 0xFF c - 0
// Black: bw - 0 c - 0
// Red: bw - 0xFF c - 0xFF
```

@caemor could you help with extending `TriColor` implementation to match these requirements? I am willing to do it myself, let's just agree on the best way to approach it. Thanks!

### DIFF
--- a/examples/epd4in2_variable_size.rs
+++ b/examples/epd4in2_variable_size.rs
@@ -70,7 +70,8 @@ fn main() -> Result<(), std::io::Error> {
     let (x, y, width, height) = (50, 50, 250, 250);
 
     let mut buffer = [epd4in2::DEFAULT_BACKGROUND_COLOR.get_byte_value(); 62500]; //250*250
-    let mut display = VarDisplay::new(width, height, &mut buffer, false).unwrap();
+    let mut display =
+        VarDisplay::new(width, height, &mut buffer, DisplayMode::BwrBitOff as u8).unwrap();
     display.set_rotation(DisplayRotation::Rotate0);
     draw_text(&mut display, "Rotate 0!", 5, 50);
 
--- a/src/color.rs
+++ b/src/color.rs
@@ -3,6 +3,7 @@
 //! EPD representation of multicolor with separate buffers
 //! for each bit makes it hard to properly represent colors here
 
+use crate::prelude::DisplayMode;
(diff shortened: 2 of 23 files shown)


yurinnick:
After this modification colors render correctly. I need to look how to integrate it with current implementation.
```
impl ColorType for TriColor {
    const BITS_PER_PIXEL_PER_BUFFER: usize = 1;
    const BUFFER_COUNT: usize = 2;

    fn bitmask(&self, bwrbit: bool, pos: u32) -> (u8, u16) {
        let bit = 0x80 >> (pos % 8);
        let mask = match self {
            TriColor::Black => (bit as u16) << 8,
            TriColor::White => (bit as u16) << 8 | bit as u16,
            TriColor::Chromatic => 0u16,
        };
        (!bit, mask)
    }
}
```

yurinnick:
Also, `update_partial_frame()` doesn't work for whatever reason.

simon0356:
Hello Yurinnick,
I was working on an implementation on the same screen and got the same issue about the chromatic buffer which is inverted.
I have tested your implementation and can confirm it is working on EPD4IN2bc screen 
As a reference i use the EPD_4in2b_V2 example code from Waveshare here : https://github.com/waveshare/e-Paper

Thanks for your work, 
Simon

struckoff:
Sup guys, any plans on releasing this one?

yurinnick:
@caemor could you provide feedback please?

piedoom:
Is there any additional work that needs to be done for this to be merged in, or any help needed? I would love to use this on main.

vojty:
I rebased and fixed this PR which I use for my project and created another one here https://github.com/caemor/epd-waveshare/pull/231

PR #138, opened 2022-12-07
Use feature gate to add 2in13 v3

caemor:
A different approach to #126, which removes the code duplication.

PR #145, opened 2023-02-22
Add support for Epd2in13b v4

tommy-gilligan:
I saw https://github.com/caemor/epd-waveshare/pull/138 . I'm not too sure whether V4 makes sense as a separate driver or not.  I did try getting V2 driver working with V4 display with as few as changes as possible.  As far as I can tell, there are a few more changes involved for V3->V4 than there is V2->V3 but it doesn't rule out a shared driver.

I added some tests to help me understand TriColor better (but I don't know if they're super necessary).


romarioferrao:
Hey @tommy-gilligan thanks for the work you did here. I just have a quick question. Does this code also work for waveshare 2in13 v4 (the black and white version)? Because I see there is not support for that particular device in [edp-waveshare](https://github.com/caemor/epd-waveshare).

PR #107, opened 2022-04-16
WIP: Epd2in66 module

Lite5h4dow:
this is not the final version. just wan to get this here so you can take a look and advise on any further changes, stuck on the rotation functions atm.

PR #147, opened 2023-02-27
Add pi hat Pico ePaper 2.66 B/W/R

ReinoutHeeck:
I had my own driver code for this hat but lacked graphics support. The quickest route seemed to integrate with this project - so here it is. I am aware there is another effort in WIP for this display, but that seems dormant (see #107). I needed running code 'now', therefor I used the code that I knew.

In the documentation I included a full,working example for the Pi Pico that is targeted by this hat, but this driver is platform agnostic just as all the others here.


caemor:
Thanks!

PR #148, opened 2023-02-27
set_pixel() -- fix off-by-one bug in bounds check

ReinoutHeeck:
Title says it all :-)


caemor:
Thanks, sorry for the late reply

caemor (approved):

Issue #152, opened 2023-04-25
Problems with Pico-ePaper-7.5-B


patrickelectric:
I'm using: https://www.waveshare.com/wiki/Pico-ePaper-7.5-B

I'm getting a red screen when using `epd7in5_v3`, but from what I could understand, I should be using `epd7in5b_v2`, but my esp32-s2 fails to even start when including it with the same code!

![image](https://user-images.githubusercontent.com/1215497/234167257-893fd3d6-6dfe-4855-bb8b-81b6a2ceb914.png)

patrickelectric:
it appears that the problem only happens when compiling in debug mode.

caemor:
Buffer size fix got merged in 0.6. So I'm closing this issue for now, since you got it working with the fix using the release mode.

PR #153, opened 2023-04-25
epd7in5b_v2: Fix buffer size

patrickelectric:
Fix Helps #152 
At least the code runs now, the buffer size was allocated with the wrong size


patrickelectric:
I got it working in the end when building in release mode.

PR #156, opened 2023-05-01
Fixed init code for epd1in54_vs

### LINKED ISSUE #154: epd1in54_v2 does not show anything until LUT is set.
It seems to me that the `init` function for epd1in54_v2 does forget to set the lut (Look up table), which means that code does not do anything at all unless epd.set_lut is explicitly called later.

At least my Raspberry pi pico did not draw any of the data i send until I eventually tried calling set_lut.

Sadly, I don't have any experience with ePaper displays and I don't fully understand why this is needed:
The [Reference code (4.2)](https://www.waveshare.com/w/upload/e/e5/1.54inch_e-paper_V2_Datasheet.pdf) mentions:
```
setting: Load LUT from OTP
SPI(0x22, 0xb1)
```

I guess this means init should call ``cmd_with_data(spi, Command::DisplayUpdateControl2, &[0xb1])?;`` somewhere?

Should I just try to get it working on my hardware and open a PR if my init does work in a way that the next display_frame-call does something? I don't fully understand what I am doing, but at least it would work then?

If anyone else tried the epd1in54_v2 and did not neet to set_lut explicitly, that would probably mean my software is buggy somewhere.

fakusb:
Added loading of look-up-table to init for 1in54_v2. Closes #154 .

### DIFF
--- a/src/epd1in54_v2/mod.rs
+++ b/src/epd1in54_v2/mod.rs
@@ -71,9 +71,6 @@ where
 
         self.set_ram_area(spi, delay, 0, 0, WIDTH - 1, HEIGHT - 1)?;
 
-        self.interface
-            .cmd_with_data(spi, Command::BorderWaveformControl, &[0x1])?;
-
         self.interface.cmd_with_data(
             spi,
             Command::TemperatureSensorSelection,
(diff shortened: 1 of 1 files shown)


caemor:
I can't test this, but the change seems reasonable to me.

PR #157, opened 2023-05-03
Add support for EPD 2in7

andrej-mk:
Based on python library
https://github.com/waveshare/e-Paper/blob/master/RaspberryPi_JetsonNano/python/lib/waveshare_epd/epd2in7.py


caemor:
Thanks!

caemor:
Updated and merged in #216. Thanks for your work!

PR #159, opened 2023-06-10
Adds support for the Waveshare 5.83" E-Ink Display

Carbonhell:
This module is for this display: https://www.waveshare.com/5.83inch-e-paper.htm
Based on the C library here: https://github.com/waveshare/e-Paper/blob/master/RaspberryPi_JetsonNano/c/lib/e-Paper/EPD_5in83_V2.c

I have written this module since I own this particular display. I kept the partial update out since, from what I can understand, it is not supported on it: https://www.waveshare.com/5.83inch-e-paper.htm (See selection guide)

I can confirm it works on a real hardware test :)

The code itself is fairly similar to other B/W displays, I didn't have to do much other than fixing some operation codes during init.

### DIFF
--- a/src/epd5in83_v2/command.rs
+++ b/src/epd5in83_v2/command.rs
@@ -1,0 +1,134 @@
+//! SPI Commands for the Waveshare 5.83" E-Ink Display
+
+use crate::traits;
+
+/// Epd5in83 commands
+///
+/// Should rarely (never?) be needed directly.
+///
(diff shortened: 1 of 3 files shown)


Carbonhell:
From what I can see, the clippy errors in the failed actions are unrelated to the contents of this PR. If anything needs to be changed in this PR, though, I'll try to do it as soon as possible.

caemor:
Thanks!

PR #160, opened 2023-06-11
Order displays from largest size to smallest

shymega:
This makes it easier to read, and organises the supported displays in the table.


caemor (approved):
Thanks, thats a great idea

PR #167, opened 2023-08-30
Update 'B' 1.54" e-ink display, and remove 'C' display link

### LINKED ISSUE #161: Some 1.54" display links in README 404.
Hi,

I'm trying to work out if my e-ink display is supported by this library. I was looking at the 1.54" display links and two 404 for me. I've made a PR since to organize the list, so this issue might be outdated if that's merged, but the following links are 404'ing.

1.54 Inch B/W/Y (C): https://www.waveshare.com/1.54inch-e-paper-c.htm
1.54 Inch B/W/R (B): https://www.waveshare.com/product/modules/oleds-lcds/e-paper/1.54inch-e-paper-module-b.htm

The other 1.54" display link is fine.

Is there a link we should be using instead of these broken links?

Cheers.

shymega:
The 1.54" B/W/R (B) e-ink display URI has changed. However, the 1.54" B/W/Y (C) e-ink display is no longer available.

Resolves #161.


caemor:
Thanks

PR #171, opened 2023-10-08
Add support for EPD 2in9 D

wsndshx:
I obtained an e-ink screen with the model number `WFT0290CZ10`. I found that it has the same model number as the Waveshare 2.9" D E-Ink Display, so I added the device based on the information provided by Waveshare.

References:
- [Waveshare WiKi](https://www.waveshare.com/wiki/2.9inch_e-Paper_HAT_(D)_Manual)
- [2.9inch e-Paper (D) Specification](https://files.waveshare.com/upload/b/b5/2.9inch_e-Paper_%28D%29_Specification.pdf)
- [Arduino/epd2in9d/epd2in9d.cpp](https://github.com/waveshareteam/e-Paper/blob/4822c075f5df714f88b02e10c336b4eeff7e603e/Arduino/epd2in9d/epd2in9d.cpp)


wsndshx:
It appears that the error in the failed operation is due to the upgrade of embedded-graphics from version `0.7.1` to `0.8.1`, which has broken the old feature.

caemor:
Thanks

caemor (approved):

Issue #151, opened 2023-04-19
cannot compile epd-waveshare (overflow error)


tippfehlr:
Disabling default features removes the error.
```
$ cargo build
    Updating crates.io index
   Compiling epd-waveshare v0.5.0
error[E0080]: evaluation of constant value failed
  --> /home/jonathan/.cargo/registry/src/github.com-1ecc6299db9ec823/epd-waveshare-0.5.0/src/epd4in2/graphics.rs:11:18
   |
11 |     buffer: [u8; WIDTH as usize * HEIGHT as usize / 8],
   |                  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ attempt to compute `400_usize * 300_usize`, which would overflow

error[E0080]: evaluation of constant value failed
  --> /home/jonathan/.cargo/registry/src/github.com-1ecc6299db9ec823/epd-waveshare-0.5.0/src/epd5in65f/graphics.rs:11:18
   |
11 |     buffer: [u8; WIDTH as usize * HEIGHT as usize / 2],
   |                  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ attempt to compute `600_usize * 448_usize`, which would overflow

error[E0080]: evaluation of constant value failed
  --> /home/jonathan/.cargo/registry/src/github.com-1ecc6299db9ec823/epd-waveshare-0.5.0/src/epd7in5/graphics.rs:11:18
   |
11 |     buffer: [u8; WIDTH as usize * HEIGHT as usize / 8],
   |                  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ attempt to compute `640_usize * 384_usize`, which would overflow

error[E0080]: evaluation of constant value failed
  --> /home/jonathan/.cargo/registry/src/github.com-1ecc6299db9ec823/epd-waveshare-0.5.0/src/epd7in5_hd/graphics.rs:11:18
   |
11 |     buffer: [u8; WIDTH as usize * HEIGHT as usize / 8],
   |                  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ attempt to compute `880_usize * 528_usize`, which would overflow

error[E0080]: evaluation of constant value failed
  --> /home/jonathan/.cargo/registry/src/github.com-1ecc6299db9ec823/epd-waveshare-0.5.0/src/epd7in5_v2/graphics.rs:11:18
   |
[...]

PR #173, opened 2023-10-12
fix: overflow error

ldicarlo:
Currently on my computer it displays an error message: 
```
44 | const NUM_DISPLAY_BYTES: usize = WIDTH as usize  * HEIGHT as usize /8;
   |                                  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ attempt to compute `800_usize * 480_usize`, which would overflow
```
as said in #151.
This a fix to this error (when compiling for arduino, in my case).

### DIFF
--- a/src/epd7in5b_v2/mod.rs
+++ b/src/epd7in5b_v2/mod.rs
@@ -41,7 +41,7 @@ pub const HEIGHT: u32 = 480;
 /// Default Background Color
 pub const DEFAULT_BACKGROUND_COLOR: TriColor = TriColor::White;
 
-const NUM_DISPLAY_BYTES: usize = WIDTH as usize * HEIGHT as usize / 8;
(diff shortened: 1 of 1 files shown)


tippfehlr:
What target are you compiling for? (which arduino?)

My problem as in #151 was that the buffer doesn’t fit in my memory (of an Arduino Nano) at all.

ldicarlo:
Ok. Mine is an Arduino Uno. But I may have complety misunderstood the whole config. 
I use `arduino-hal` and `ravedude` but I am kind of lost between all the tools I am using. 

What I know is that when I git clone this repo and use it as a dependency, it compile only with this version. In the master version, I get the same error as you in #151

tippfehlr:
I also use ardino-hal and ravedude (I don’t know if this even could be run with another hal or if there exists another good one)

You shouldn’t need to clone the repo for normal use though. Just add it with `cargo add epd-waveshare` (except when you are developing on it ofc)

caemor:
Thanks!

Issue #164, opened 2023-07-30
Support for 12.48inch module


barafael:
Is support for [this waveshare 12.48 inch module](https://www.waveshare.com/12.48inch-e-paper-module-b.htm) feasible, or planned?

caemor:
It should be feasible, but it's not something I've planned as I don't have one of these in my hands.

barafael:
I have one. It works fine with the C example code from waveshare, but yeah that's not great. I can either help test, or, if you like, I can send it to you. No obligations other than sending it back to me when you don't want it anymore, and only if it didn't break along the way :D let me know what you think, email is on my profile. I can send to a Packstation ofc.

barafael:
OK. I'll let you know when I need help, could be a while. Thanks!

vadimcn:
Hi!  
I have the display and I wanted to take a stab at implementing the diver.   

However... I immediately ran into conceptual difficulties with this crate's traits, because they [assume that all displays have the same interface](https://github.com/caemor/epd-waveshare/blob/main/src/interface.rs#L5), and take a fixed number of pins as generic parameters.   Well, as it turns out, epd12in48(B) actually consists of [4 sub-displays](https://www.waveshare.com/wiki/12.48inch_e-Paper_Module), each with its own CS and BUSY pins, while DC and RST pins are by 2 sub-displays).  

@caemor, would you have any advice as to how to fit this sort of device into your framework?  

The only thing that comes to mind, is exposing it as 4 separate displays.  Which might have to be done anyways, since my ESP32 driver board [apparently cannot allocate the contiguous 160K bufer](https://www.esp32.com/viewtopic.php?f=19&p=116093), that would be needed for a 1304*984 image.

PR #174, opened 2023-10-23
Add support for epd12in48b_v2

vadimcn:
Here's a first draft of my driver for [12.48" module](https://github.com/caemor/epd-waveshare/issues/164).  Does this look like something you'd be willing to integrate?

### DIFF
--- a/src/epd12in48b_v2/mod.rs
+++ b/src/epd12in48b_v2/mod.rs
@@ -1,0 +1,679 @@
+//! A driver for the Waveshare 12.48"(B) E-Ink Display (V2) via SPI
+//! (also known as [GDEY1248Z51](https://www.good-display.com/product/422.html))
+//!
+//! # References
+//!
+//! - [Datasheet](https://files.waveshare.com/upload/b/b4/12.48inch_e-Paper_B_V2_Specification.pdf)
+//! - [Wiki](https://www.waveshare.com/wiki/12.48inch_e-Paper_Module_(B))
+//! - [Waveshare C drivers](https://github.com/waveshareteam/12.48inch-e-paper/)
+//!
+
+mod command;
+mod config;
+
+use embedded_hal::{
+    delay::DelayNs,
+    digital::{InputPin, OutputPin, PinState},
+    spi::SpiBus,
+};
+
+pub use crate::rect::Rect;
+use command::Command;
+pub use config::*;
+
+/// A collection of peripherals controlling the EPD
+///
+/// The display is composed of 4 sub-displays arranged like so:
+/// ```
+///     0        648      1304
+///   0 +--------+--------+
+///     |   S2   |   M2   |
+/// 492 +--------+--------+
+///     |   M1   |   S1   |
+/// 984 +--------+--------+
+/// ```
+/// Resolution of `S2` and `M1` is 648 x 492,
+/// resolution of `S1` and `M2` is 656 x 492.
+///
+pub struct Peripherals<INPUT, OUTPUT, SPI>
+where
+    INPUT: InputPin,
+    OUTPUT: OutputPin,
+    SPI: SpiBus<u8>,
+{
+    /// SPI bus shared by all sub-displays.
+    pub spi: SPI,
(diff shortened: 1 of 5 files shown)


i-like-noodles:
I have the same device and same ESP32 and this worked for me. But mine is v1 because it did not come with a v2 sticker as the [documentation](https://www.waveshare.com/wiki/12.48inch_e-Paper_Module_(B)#Version_Guide) says. This device has a slightly different startup procedure in the upstream code [as seen here](https://github.com/waveshareteam/12.48inch-e-paper/blob/558effea7955b082d9ebe01ed2432cf0c2fba7b8/esp32/esp32-epd-12in48/src/utility/EPD_12in48b.cpp#L69), though I'm not sure if it makes a difference. The only thing it seems to change is that the v2 startup flashes a lot more compared to v1 on mine.

I did a quick change to this to implement v1 as well, though I don't know if this is how you would want to implement it. See this [diff.patch](https://github.com/caemor/epd-waveshare/files/13208693/diff.patch)

src/epd12in48b_v2/rect.rs
caemor, line 1: We should move this to a more unspecific module (e.g. helper or directly at the top level below lib.rs), since it's nothing really specific to the epd12in48b_v2

caemor (changes requested):
Thanks for your work!

I would prefer to have the example in your own, separate repository, which we can link to from this repository.

The rest looks good to me.

caemor:
> I have the same device and same ESP32 and this worked for me. But mine is v1 because it did not come with a v2 sticker as the [documentation](https://www.waveshare.com/wiki/12.48inch_e-Paper_Module_(B)#Version_Guide) says. This device has a slightly different startup procedure in the upstream code [as seen here](https://github.com/waveshareteam/12.48inch-e-paper/blob/558effea7955b082d9ebe01ed2432cf0c2fba7b8/esp32/esp32-epd-12in48/src/utility/EPD_12in48b.cpp#L69), though I'm not sure if it makes a difference. The only thing it seems to change is that the v2 startup flashes a lot more compared to v1 on mine.
> 
> I did a quick change to this to implement v1 as well, though I don't know if this is how you would want to implement it. See this [diff.patch](https://github.com/caemor/epd-waveshare/files/13208693/diff.patch)

It would be great to support v1 as well. I think we could solve this with feature-gating without much code duplication.

src/rect.rs
```diff
@@ -0,0 +1,87 @@
+//! Rectangle operations
```
caemor, line 1:
```suggestion
//! Rectangle operations for bigger displays with multiple _windows_
```

src/epd12in48b_v2/mod.rs
```diff
@@ -0,0 +1,679 @@
...
+};
+
+pub use crate::rect::Rect;
+use command::Command;
+pub use config::*;
+
+/// A collection of peripherals controlling the EPD
+///
+/// The display is composed of 4 sub-displays arranged like so:
+/// ```
```
caemor, line 27:
```suggestion
/// ```md
```

caemor (approved):

ldicarlo:
Hello, I don't want to open an issue for that since you advised @vadimcn to do it in his repository, but, @vadimcn would you mind, please, show me an example of how to use the driver you wrote here? 
I am not sure that it should be called like other drivers here, since it has 4 screens. 
Thank you for your work, both @caemor  and @vadimcn !

vadimcn:
Here's my test driver app: https://github.com/vadimcn/epd-waveshare/tree/main/esp32/src

ldicarlo:
thank you for you prompt answer. I still have a question if you don't mind. 
Would you think that your driver would work also with a raspberry ? I see that I need to instanciate Spi and Pins differently than esp's but I am not sure the code could work. 
I'll adapt your example as I can

vadimcn:
> Would you think that your driver would work also with a raspberry?

Haven't tried it, but I don't see why not.  That's the whole point of embedded-hal.

PR #176, opened 2023-11-06
fix build when feature graphics is not enabled

vhdirk:
`PixelColor` is only available when `embedded_graphics` is used


caemor:
Thanks

PR #179, opened 2023-11-29
Fix Examples

DaveRichmond:
So the underlying linux-embedded-hal renamed Spidev to SpidevDevice. Sensical change, but the examples needed to be updated to match that change.

I've only tested that the examples build again, I don't have easy access to hardware to test the results on right now


caemor:
Thanks!
This got fixed by the #181

PR #187, opened 2024-01-22
Use async embedded-hal

newAM:
This is not complete, and I am not going to be actively working on it (not for a few months anyway).  I wanted to share an up-to-date version since #175 is out of date, and also no longer being worked on.

Things to fix:

1. waiting for the busy pin could be implemented better.
2. examples do not compile, `linux-embedded-hal` does not implement async
3. synchronous support

Anyone please feel free to use any commits here in other pull-requests.


vhdirk:
Did you see this one? https://github.com/vhdirk/epd-waveshare-async

PR #191, opened 2024-02-04
feat: add support for 7.3 Inch HAT (F)

jetjinser:


src/color.rs
```diff
@@ -44,7 +44,7 @@ pub enum TriColor {
     Chromatic,
 }
 
-/// For the 5in65 7 Color Display
+/// For the 7 Color Display
```
caemor, line 47:
```suggestion
/// For the 5in65 and 7 Color Display
```

caemor (changes requested):
Looks great. Thanks!

src/color.rs
caemor:
```suggestion
/// For the 7 Color Displays
```

PR #196, opened 2024-03-11
Update to work with latest e-g and tinybmp

bemyak:
I had some issues with rendering simple black & white images with tinybmp crate, and had to apply these patches to make it work. Let me know if it makes sense.

### DIFF
--- a/src/color.rs
+++ b/src/color.rs
@@ -301,6 +301,13 @@ impl From<embedded_graphics_core::pixelcolor::raw::RawU1> for Color {
 }
 
 #[cfg(feature = "graphics")]
+impl From<Color> for embedded_graphics_core::pixelcolor::raw::RawU1 {
+    fn from(color: Color) -> Self {
+        Self::new(color.get_bit_value())
+    }
+}
+
+#[cfg(feature = "graphics")]
 impl PixelColor for Color {
     type Raw = embedded_graphics_core::pixelcolor::raw::RawU1;
 }
--- a/src/color.rs
+++ b/src/color.rs
@@ -346,8 +346,68 @@ impl From<Color> for embedded_graphics_core::pixelcolor::Rgb888 {
     fn from(color: Color) -> Self {
         use embedded_graphics_core::pixelcolor::RgbColor;
         match color {
-            Color::Black => embedded_graphics_core::pixelcolor::Rgb888::BLACK,
-            Color::White => embedded_graphics_core::pixelcolor::Rgb888::WHITE,
+            Color::Black => embedded_graphics_core::pixelcolor::Rgb888::WHITE,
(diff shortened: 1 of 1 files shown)


avlec:
I've had some success on the 3.7in (which doesn't have the best support) without any modifications. Just wondering if this change makes working with these slightly easier (no color_converted() in the draw call?)
```rust
use embedded_graphics::pixelcolor::BinaryColor;
use embedded_graphics::{prelude::*, image::Image};

// ...
    let mut display = Display3in7::default();
    const SPLASH: &[u8] = include_bytes!("../splashnewbw.bmp");
    let splash_bmp = tinybmp::Bmp::<BinaryColor>::from_slice(SPLASH).unwrap();
    Image::new(&splash_bmp, Point::zero()).draw(&mut display.color_converted()).unwrap_or_else(|_| error!("failed to render splash screen"));
```

diajowe:
I see this PR is a draft? Is there something missing for completion?

bemyak:
Yeah, there were some arguable changes related to color inversion, so I was waiting to get any feedback.

I've dropped them for now, and users are encouraged to use `display.color_converted()` as mentioned by @avlec.

I'll mark this as ready for review now.

diajowe:
> Yeah, there were some arguable changes related to color inversion, so I was waiting to get any feedback.
> 
> I've dropped them for now, and users are encouraged to use `display.color_converted()` as mentioned by @avlec.
> 
> I'll mark this as ready for review now.

Would you mind going a bit more into detail on what could go wrong?
Maybe I could test a bit further, as currently, I only displayed some data using 7.5inch BW version of waveshare EPD

bemyak:
> Would you mind going a bit more into detail on what could go wrong?

`BinaryColor` is supposed to represent something like a LED pixel, so `BinaryColor::On` means the LED is on, thus the value should (in my opinion) correspond to `Color::White` on an EPD.

However, in the current implementation, it's the opposite: https://github.com/caemor/epd-waveshare/blob/1244f03591bcc5f001b6e634170eb99e29e768d0/src/color.rs#L297-L304

I _guess_ it's because in most cases when rendering text, people expect to see white on black for LED panels and black on white for EPD. Displaying images becomes tricky now since you need to invert colours to display them properly.

Inverting all the colours to be "On=White, Off=Black" is a breaking change and it must be discussed thoroughly.

caemor:
Sorry for the late response.

I followed @andber1 suggestion for PixelColor for Color and TriColor in https://github.com/caemor/epd-waveshare/pull/215
I would prefer that this PR follows in the same direction and I think only the conversion from 565 is missing now.

bemyak:
I've updated the code to only add Rgb565, Rgb555 the missing Color → RawU1 conversions

caemor:
Thanks a lot.

caemor (approved):

PR #209, opened 2024-10-13
Add support for epd1in02

brrastak:
Fully tested and works


caemor:
Thanks for your work!

PR #214, opened 2024-10-29
Add update_and_display_frame to example in readme

andber1:
I think the example on the readme is missing one important step: actually showing the display on the e-paper.


caemor:
I completely agree with you. Thanks for the fix.

PR #192, opened 2024-02-17
Add example 7in5 v2 program

blamster19:
The program tests rotations and draws analog clock (based on example code for different models from the repository), tests default fonts of embedded-graphics crate and displays an image of Ferris from `examples/assets/` directory.

I thought it would be nice to show how to display an image file, because I myself spent too much time fiddling with wrong types (I found solution to this problem in #130).

The program has been tested on hardware and works as expected.

PR #215, opened 2024-10-29
Implement PixelColor for Color and TriColor

andber1:
`OctColor` already has an implementation for `PixelColor`. I added implementations for `Color` and `TriColor`, so that one can draw images without using `color_converted()`.
The documentation of `color_converted` says, it should be avoided (see [here](https://docs.rs/embedded-graphics/latest/embedded_graphics/draw_target/trait.DrawTargetExt.html#tymethod.color_converted)). So I think this might be helpful.


andber1:
@blamster19 FYI: I noticed you "spent much time fiddling with wrong types" in #192. You may find this interesting.

caemor:
Thanks!!

PR #219, opened 2024-10-30
Fix update_old_frame for epd2in9_v2

andber1:
The problem was introduced in #103 and leads to some pixels not updated in the next quick refresh after a full refresh. The function `update_old_frame` now does the same thing as `SetFrameMemory_Base`in the original C code [here](https://github.com/waveshareteam/e-Paper/blob/af4d8b49ccef5f8f5fb88e9a836b86bd3f0bbfe3/Arduino/epd2in9_V2/epd2in9_V2.cpp#L425C11-L425C30).


caemor (approved):

caemor:
Thanks for the bugfix!

PR #230, opened 2025-01-19
Optional async support

XLPhere:
This is an attempt to build upon PR #175 by @vhdirk with the idea by @avsaase of using the [maybe_async](https://github.com/fMeow/maybe-async-rs).

The `async` feature will do the following:
- Switch all the relevant functions from sync to async.
- Swap `DelayNs` and `SpiDevice` from `embedded_hal`  to their `embedded_hal_async` counterparts.

The actual implementation code is written in async, but will be converted to sync code by `maybe_async` if the `async` feature is not defined, so this should not result in any runtime overhead and code duplication should be minimal.

Notes:
- This solution uses a modified version of  the `maybe_async` crate to make the async feature opt in and not opt-out, see: https://github.com/fMeow/maybe-async-rs/pull/26
- `#[maybe_async::maybe_async(AFIT)]` currently doesn't have a possibility make returned version `Send`, possible solutions to this might be:
  1. Ignore the compiler warning with `#[allow(async_fn_in_trait)]` and not guarantee the presence of the Send trait. (current solution used in this draft)
  2. Use `#[maybe_async::maybe_async(Send)]`, which will use `#[async_trait]` (in its current version) - would be an additional dependency added.
  3. Define sync and async version of crates manually (would duplicate trait definition).
- The slowdown from `SINGLE_BYTE_WRITE = true` is significantly higher on async (for the RP2040 I'm testing on at least, it slows down to 40s) - a feature to disable this behavior would possibly be beneficial (or setting this to false entirely).

Comments and improvement ideas are appreciated!

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -27,6 +27,8 @@ jobs:
       run: cargo build --examples --all-targets --verbose
     - name: Run tests
       run: cargo test --verbose
+    - name: Run tests (async)
+      run: cargo test --verbose --lib --tests --benches --features async
     - name: Build docs
       run: cargo doc
     - name: Clippy
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -16,7 +16,9 @@ rust-version = "1.62"
 [dependencies]
 embedded-graphics-core = { version = "0.4", optional = true }
 embedded-hal = "1.0.0"
+embedded-hal-async = { version = "1.0.0", optional = true }
 bit_field = "0.10.1"
+maybe-async = { git = "https://github.com/XLPhere/maybe-async-rs.git", branch = "default_sync", features = ["default_sync"] }
 
 [dev-dependencies]
 embedded-graphics = "0.8"
@@ -47,11 +49,16 @@ required-features = ["linux-dev"]
 name = "epd4in2"
 required-features = ["linux-dev"]
 
+[[example]]
+name = "epd7in5_v2"
+required-features = ["linux-dev"]
+
 [features]
 # Remove the linux-dev feature to build the tests on non  unix systems
 default = ["graphics", "linux-dev", "epd2in13_v3"]
 
 graphics = ["embedded-graphics-core"]
+async = ["embedded-hal-async", "maybe-async/is_async"]
 epd2in13_v2 = []
 epd2in13_v3 = []
 linux-dev = []
--- a/src/epd12in48b_v2/mod.rs
+++ b/src/epd12in48b_v2/mod.rs
@@ -11,12 +11,9 @@
 mod command;
 mod config;
 
-use embedded_hal::{
-    delay::DelayNs,
-    digital::{InputPin, OutputPin, PinState},
-    spi::SpiBus,
-};
+use embedded_hal::digital::{InputPin, OutputPin, PinState};
 
+use crate::interface::{DelayNs, SpiBus};
 pub use crate::rect::Rect;
 use command::Command;
 pub use config::*;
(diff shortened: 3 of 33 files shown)


XLPhere:
This should now be ready for testing!
Feedback on the async version as well as the sync version (which should work exactly the same as before) is very much appreciated!

vhdirk:
Hi @XLPhere . 
While it may not look like it, my async fork of this project does work. Or did work, I can't remember where I left off trying to go for embedded hal 1.0 :) I have a number of other projects queued before I can invest time in this again, so I'm glad you are taking the time getting async merged.

Maybe_async is cool. It keeps API changes (for sync) to a minimum, if needed at all. My idea was to work on an async version first and then see how that could be merged into main again.

Specifically, I wanted to get rid of all polling that exists in epd-waveshare since embedded_hal_async has a Wait trait. Even if rusts async is all polling, I felt that using the built-in Wait would perhaps be beneficial to battery life. But that's just an assumption. 

As for delays; I saw that spi has a 'delay' transaction. I am assuming it to be more optimized than just a regular delay, but that's not really based on anything. 

Since those 2 things caused a lot of API changes, I tried to improve error handling at the same time. 

And as these things go, it ultimately became too much work to chop it in bite-sized chunks for merging...

I hope this clears up what my fork is trying to achieve :)

cvdbcd:
Hey @vhdirk

Thank you for your comments and previous work on this that very much influenced this solution!

The idea of this PR is to introduce no API and essentially also no implementation changes at all for the sync version, so the existing users won't run in any trouble from upgrading to a version with these changes.

The idea of the Wait trait is a good one, I have seen that some implementations use interrupts to facilitate that instead of polling, so it might make a difference. I will look into a solution that will use the wait trait when it is available and no `delay_us` has been set by the user.

The idea of SPI having a delay transaction could be useful to simplify the API (by no longer needing to pass the delay object), but this would API changes or at least could introduce behavioral changes for the sync version, so i would consider this out of scope for this PR. If seen beneficial it can be added by a separate PR later, i would suggest.

Either way, thanks for your input!

XLPhere:
Hey @vhdirk

Thank you for your comments and previous work on this that very much influenced this solution!

The idea of this PR is to introduce no API and essentially also no implementation changes at all for the sync version, so the existing users won't run in any trouble from upgrading to a version with these changes.

The idea of the Wait trait is a good one, I have seen that some implementations use interrupts to facilitate that instead of polling, so it might make a difference. I will look into a solution that will use the wait trait when it is available and no delay_us has been set by the user.

The idea of SPI having a delay transaction could be useful to simplify the API (by no longer needing to pass the delay object), but this would API changes or at least could introduce behavioral changes for the sync version, so i would consider this out of scope for this PR. If seen beneficial it can be added by a separate PR later, i would suggest.

Either way, thanks for your input!

XLPhere:
Hi @vhdirk, I've been looking a bit into seamlessly integrating the `Wait` trait when available by using [autoderef-specialisation](https://github.com/dtolnay/case-studies/tree/master/autoref-specialization), but it sadly didn't work out as expected ([my attempt](https://github.com/XLPhere/epd-waveshare/commit/36c35a0e9ff0aa1433ef22a3f337efe8bdc62679)). So i would suggest moving that optimization to a possible future PR for somebody else to have a stab at it.

XLPhere:
@caemor do you have some info on which things still need to be addressed for this PR to be ready for merge?

fakusb:
Nice idea, I like it.

But it seems that the Busy-pin is not yet async (embedded-hal::digital::Wait), which would be useful for displays with long refresh delay. I have not found a nice solution to add the `+ Wait`-bound conditionally yet.

fakusb:
After looking deeper into it, it currently polls every 10 ms in this PR, which might be enough for most applications.

I'm really eager to hear if this kind of PR is considered for mearging.

PR #235, opened 2025-04-28
feat: add Epd7in3e based on e-ink Spectra6

RitwikSaikia:
Implementation for [7.3inch_e-Paper_HAT_(E)](https://www.waveshare.com/wiki/7.3inch_e-Paper_HAT_(E)).  which is based on [e-Ink Spectra6](https://www.eink.com/brand/detail/Spectra6) technology.

Tested on my personal display.

### DIFF
--- a/src/color.rs
+++ b/src/color.rs
@@ -46,6 +46,24 @@ pub enum TriColor {
     Chromatic,
 }
 
+/// For the 6 Color Displays
+#[derive(Clone, Copy, PartialEq, Eq, Debug, Default)]
+pub enum HexColor {
+    /// Black Color
+    Black = 0x00,
+    /// White Color
+    #[default]
+    White = 0x01,
+    /// Yellow Color
+    Yellow = 0x02,
+    /// Red Color
+    Red = 0x03,
+    /// Blue Color
+    Blue = 0x05,
+    /// Green Color
+    Green = 0x06,
+}
+
 /// For the 7 Color Displays
 #[derive(Clone, Copy, PartialEq, Eq, Debug, Default)]
 pub enum OctColor {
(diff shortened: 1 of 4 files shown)


RitwikSaikia:
What is required for this PR to be merged?

dragazo:
Does this support full color? It looks like it just uses 6 colors unless I missed something. Also, it looks like some functions still have `unimplemented!()` placeholders, which will cause a firmware panic.

ozwaldorf:
> Does this support full color? It looks like it just uses 6 colors unless I missed something. 

@dragazo The waveshare "full-color" displays are only using 6 colors. However, the colors are very vibrant so using dithering you can get a pretty decent looking image. But unfortunately it's not supporting full rgb.

ozwaldorf:
Also having done some research into the board, it seems it doesn't support custom lut waveforms so the set_lut trait item wouldn't ever work for this display.

Similarly, there's no documentation from the manufacturer on how to send a partial or windowed update, so it seems only full buffer updates are possible

ozwaldorf:
I have tested this using the ESP32-S3-PhotoFrame and can confirm the basic driver functionality is working as expected.

ozwaldorf:
After reaching out to a few manufacturers producing this display, I managed to find a working implementation of partial refreshes! Additionally, I ported the sample into rust and successfully ran the partial refresh on the waveshare panel and driver.

The sample code is from the GoodDisplay variant, in the downloads section: https://www.good-display.com/product/533.html

## Partial Update Implementation

Working for both GDEP073E01 or Waveshare Epd7in3e (Spectra 6e)

### Command

`0x83` (PTLW) - Partial Window command with 9 data bytes:

| Byte | Description |
|------|-------------|
| 0 | X start MSB (bits 9:8) |
| 1 | X start LSB (bits 7:0) |
| 2 | X end MSB (bits 9:8) |
| 3 | X end LSB (bits 7:0) |
| 4 | Y start MSB (bits 9:8) |
| 5 | Y start LSB (bits 7:0) |
| 6 | Y end MSB (bits 9:8) |
| 7 | Y end LSB (bits 7:0) |
| 8 | `0x01` (enable partial mode) |

Coordinates are 10-bit values (max 1023). End coordinates are inclusive.

### Sequence

  1. Set partial window (`0x83`) with coordinates
  2. Wait for BUSY
  3. Send DTM (`0x10`) with pixel data for the region only
  4. Wait for BUSY
  5. Power on (`0x04`)
  6. Wait for BUSY
  7. Set booster (`0x06`) - same values as full refresh
  8. Refresh (`0x12`, `0x00`)
  9. Wait for BUSY
  10. Power off (`0x02`, `0x00`)
  11. Wait for BUSY

### Notes

  - X and width must be even (byte-aligned, 2 pixels per byte)
  - Pixel format unchanged: 4bpp packed (same as full refresh)
  - Buffer size = `(width × height) / 2` bytes
  - Uses same waveform as full refresh (no "fast partial" mode) - just updates smaller region
  - Display must be initialized/awake before partial update

rsov:
Tested it on [reTerminal E1002](https://www.seeedstudio.com/reTerminal-E1002-p-6533.html)

Not sure if a bug with `show_6block` but the colors are correct compared to the `epd7in3f` driver

(sorry for washed out picture)

Screenshot_20260208_173324

ozwaldorf:
@rsov by default the extra framebuffer space is filled with random garbage, the blocks only write a portion, worked the same for me. Tbh would be better to display a 3x2 grid for each color instead

PR #249, opened 2025-12-17
feat: add support for partial refresh and setting LUT epd7in5_v2

dexterdy:
Good evening,

I could not get the demo to work for the epd7in5_v2 on my rpi. After a lot of fiddling, I gave up and set about to manually transpile the C demo to rust code, since that one did work. Took me some time, got it working, then started making it idiomatic rust. Then came onto the same issues as I had with the demo here. I managed to solve it. Turns out: you don't need to manually set the CS pin (that is done for you), but you DO need to set the power pin. Without that it won't work. Took the liberty to update the newer CdevPin api as well.

Anyway, I now had a bunch of rust code dealing with raw hexadecimal to drive the display and I noticed that the init functions were all incredibly different. Specifically, the C library has a different init function for full, fast, partial and grasyscale refreshing. Some had power and booster settings, others touched more esoteric stuff. None were setting anything related to refresh modes or LUT.

After a some experimenting, I figured out which commands specifically were responsible for making the refresh modes work and which commands were superfluous. 

Apparently, the manufacturers are using a hack: using the temperature override command to select a custom LUT in the OTP memory (better explanation in the code).

I took the liberty of using this knowledge to implement the set_lut function and update_partial_frame functions. 

There is also an incredibly cursed hack in the waveshare C libraries that uses the OLD/NEW buffers in the display combined with the temperature override LUT selector hack to fool the display into displaying grayscale (4 shades), which is a feature the website claims. I have not implemented this, as I think it could be harmful for the device itself.

Here a small changelog:

- fixed and updated demo for epd7in5_v2
- implemented `set_lut` for epd7in5_v2
- added `PartialRefresh` as a mode for `RefreshLut`
- implemented `update_partial frame` for epd7in5_v2

I hope you will consider my contribution.


dexterdy:
Some notes about this PR:

- I added a new struct (PartialFrame), with the purpose to make it easier to work with partial updates on devices that require the partial window to be byte-aligned in the full display. This PartialFrame translates operations on the arbitrarily places virtual partial frame, onto a byte-aligned buffer. It also takes care to keep the full display buffer in sync, and to fill the padding bits with data from the full display, so that they won't be replaced with a random empty column.
- The partial frame does depend on the Vec from the alloc crate, to make a byte-aligned buffer
- I changed the code in the epd_7in5_v2 driver to wait until idle _after_ each call that requires waiting, instead of _before_ every single call. This seems cleaner and more self-contained

src/epd7in5_v2/mod.rs
```diff
@@ -128,29 +130,65 @@ where
...
     ) -> Result<(), SPI::Error> {
-        unimplemented!();
+        let expected_size = buffer_len(width as usize, height as usize);
+        let actual_size = buffer.len();
+        if actual_size != expected_size {
+            panic!("Buffer is incorrect size. Expected: {expected_size}. Actual: {actual_size}.")
+        }
+
+        let x_aligned = x & !0b111; // force to 8-bit-boundary
+        let x_end = x_aligned + width - 1;
```
mjonuschat: This uses x_aligned (which is <= x) as the base, making the partial window narrower than intended when x is not byte-aligned. Should this be x + width - 1 instead?

When width is 0, x + width - 1 evaluates to x - 1, making the window end before the start, an early return for zero dimensions make sense.

src/epd7in5_v2/mod.rs
```diff
@@ -128,29 +130,65 @@ where
...
+        spi: &mut SPI,
+        delay: &mut DELAY,
+        buffer: &[u8],
+        x: u32,
+        y: u32,
+        width: u32,
+        height: u32,
     ) -> Result<(), SPI::Error> {
-        unimplemented!();
+        let expected_size = buffer_len(width as usize, height as usize);
```
mjonuschat, line 157: If I see this correctly, the partial window is byte-aligned (wider than width), so the actual data consumed matches the aligned width. A caller providing an aligned buffer (e.g., from get_update_parameters) could have buffer.len() > buffer_len(width, height) and hit the exact-match panic. 

I think this should validate against the aligned width, and change != to < to accept oversized buffers.

src/epd7in5_v2/mod.rs
```diff
@@ -128,29 +130,65 @@ where
...
+        let x_aligned = x & !0b111; // force to 8-bit-boundary
+        let x_end = x_aligned + width - 1;
+        let x_end_aligned = x_end | 0b111; // exclusive end boundary, ending with 3 1's (following spec)
+
+        let hrst_upper = (x_aligned >> 8) as u8;
+        let hrst_lower = x_aligned as u8;
+        let hred_upper = (x_end_aligned >> 8) as u8;
+        let hred_lower = x_end_aligned as u8;
+
+        let y_end = y + height - 1;
```
mjonuschat, line 170: Same concern about height = 0 here

src/graphics.rs
```diff
@@ -308,6 +354,305 @@ impl VarDisplay<'_, TriColor> {
...
+        y: u32,
+        width: u32,
+        height: u32,
+        full_display_buffer: &'a mut [u8],
+        full_display_width: u32,
+        full_display_size: usize,
+        bwrbit: bool,
+    ) -> Self {
+        let aligned_x = x & !0b111;
+        let x_end = x + width - 1;
```
mjonuschat, line 490: Same concern about width/heigh being zero as in the epd7in5b_v2 driver.

src/graphics.rs
```diff
@@ -308,6 +354,305 @@ impl VarDisplay<'_, TriColor> {
...
+        }
+    }
+}
+
+impl<'a, COLOR: ColorType + PixelColor> PartialFrame<'a, COLOR> {
+    /// Creates a byte-aligned buffer for you, based on X-coordinate and height.
+    ///
+    /// Parameters are documented in `Display` as they are the same as the const generics there.
+    /// bwrbit should be false for non tricolor displays
+    fn new(
```
mjonuschat, line 467: This constructor is missing a bounds check, copy_and_sync_buffer (called  from get_update_parameters) can index out of bounds on the full display buffer if the region extends past the display edge

src/graphics.rs
```diff
@@ -308,6 +354,305 @@ impl VarDisplay<'_, TriColor> {
...
+        // Apply offset to the appropriate virtual coordinate
+        match self.rotation {
+            DisplayRotation::Rotate0 => {
+                pixel.0.x += diff; // Add to X
+            }
+            DisplayRotation::Rotate90 => {
+                pixel.0.y += diff; // Add to Y
+            }
+            DisplayRotation::Rotate180 => {
+                pixel.0.x -= diff; // Subtract from X
```
mjonuschat: Shouldn't the offset be right_padding_bits instead of diff? Subtracting diff seems to shift in the wrong direction

src/graphics.rs
```diff
@@ -308,6 +354,305 @@ impl VarDisplay<'_, TriColor> {
...
+/// is replaced by a `new` method.
+pub struct PartialFrame<'a, COLOR: ColorType + PixelColor> {
+    original_x: u32,
+    aligned_x: u32,
+    y: u32,
+    original_width: u32,
+    aligned_width: u32,
+    height: u32,
+    bwrbit: bool,
+    buffer: Vec<u8>,
```
mjonuschat: This requires extern crate alloc, which isn't alwyas available in no_std environments without a heap allocator. Accepting a caller-provided &mut [u8] slice (with a required_buffer_size() helper to compute the needed size) could be a more idiomatic no_std-compatible alternative.

src/epd7in5_v2/mod.rs
```diff
@@ -197,11 +231,49 @@ where
...
+                self.cmd_with_data(spi, Command::BoosterSoftStart, &[0x27, 0x27, 0x18, 0x17])?;
+                // This selects a speed-optimized waveform: fewer voltage transitions mean faster updates
+                // (~2s vs ~4s) at the cost of increased ghosting.
+                self.cmd_with_data(spi, Command::CascadeSetting, &[0x02])?;
+                self.cmd_with_data(spi, Command::ForceTemperature, &[0x5A])?;
+            }
+            Some(RefreshLut::PartialRefresh) => {
+                // This waveform applies gentle voltage transitions that update only the changed
+                // pixels without the full-screen flicker normally required to clear ghosting.
+                // Will accumulate hosting over many cycles - requires occasional full refresh to
```
mjonuschat: Nit: Should be "ghosting" instead of "hosting"

src/epd7in5_v2/mod.rs
```diff
@@ -197,11 +231,49 @@ where
...
     fn set_lut(
         &mut self,
-        _spi: &mut SPI,
+        spi: &mut SPI,
         _delay: &mut DELAY,
-        _refresh_rate: Option<RefreshLut>,
+        refresh_rate: Option<RefreshLut>,
     ) -> Result<(), SPI::Error> {
-        unimplemented!();
+        if Some(self.refresh) == refresh_rate {
```
mjonuschat: Early return doesn't fire for `None` value. Based on treating Full and None identically below canonicalizing None to Full before the comparison might be desirable to avoid sending extra SPI commands

src/graphics.rs
```diff
@@ -308,6 +354,305 @@ impl VarDisplay<'_, TriColor> {
...
+            full_display_size,
+            rotation: DisplayRotation::default(),
+            _color: PhantomData,
+        }
+    }
+
+    /// get the number of used bytes in the buffer
+    fn buffer_size(&self) -> usize {
+        self.buffer.len()
+    }
```
mjonuschat, line 514: The suggested alloc changes would make  oversized buffers more likely,  which could in consequence have get_update_parameters returns a buffer slice that's too. Buffer size likely needs to be computed and stored.

src/epd7in5_v2/mod.rs
```diff
@@ -39,17 +39,18 @@ pub const WIDTH: u32 = 800;
 /// Height of the display
 pub const HEIGHT: u32 = 480;
 /// Default Background Color
-pub const DEFAULT_BACKGROUND_COLOR: Color = Color::White;
+pub const DEFAULT_BACKGROUND_COLOR: Color = Color::Black;
```
mjonuschat: I think this should be kept as white, the changing the VCOM setting to 0x29 fixes the polarity inversion and Black now produces actually black pixels on this display.

mjonuschat (commented):
A number of smaller things / edge cases, but I can confirm that this PR makes partial refresh on the display work.

mjonuschat (approved):

src/epd7in5_v2/mod.rs
dexterdy: The x, y, width and height parameters should correspond to the actual buffer, not to a smaller area within that buffer. That's what the check is for. The buffer shouldn't be oversized for the parameters.

I do think I should make a different change. If you pass a buffer that's not byte-aligned, the buffer will be too short for the window created by the aligned x and width, scrambling the pixels. I should write a check to ensure that the passed x and width parameters are actually aligned and then remove the code to make x and width aligned. Combined with the check for the correct buffer size, this ensures that when the code doesn't panic, the image is always drawn correctly.

dexterdy:
@mjonuschat thank you for your earlier review. I have implemented your feedback. Please consider my merge request again.

PR #254, opened 2026-04-11
feat: add support for Waveshare 3.52" EPD (EPD3IN52)

cogwheel886:
## Waveshare 3.52" E-Paper Display (EPD3IN52)

Adds driver support for the Waveshare 3.52" e-paper display (360×240px, UC8179 controller).

### Display characteristics

This display has an unusual refresh sequence not seen in other supported displays — `display_frame()` must explicitly send LUT waveform tables before triggering refresh. Most Waveshare displays handle this internally; the 3.52" does not. Additionally, the waveform tables alternate between two configurations (R22/R23) on each refresh cycle via a `lut_flag: bool` field on the driver struct.

### Hardware verification

- Tested on Raspberry Pi 5 (RP1 GPIO controller, lgpio backend)
- `IS_BUSY_LOW: true` confirmed empirically
- LUT waveform tables transcribed from Waveshare Python reference driver and verified against hardware
- Init sequence register values verified against hardware

### What's included

- `command.rs`: 17 SPI commands for the UC8179 controller
- `constants.rs`: 10 LUT waveform tables (5× GC global clear, 5× DU differential update)
- `mod.rs`: Full driver implementation with alternating waveform logic
- `Display3in52` type alias under `#[cfg(feature = "graphics")]`
- 2 new tests: `epd_size` and `command_addr`
- All 56 tests passing (54 original + 2 new)

### Not yet implemented

- `update_partial_frame()` — marked `todo!()`, hardware validation pending
- DU (differential update) LUT path — tables are present in `constants.rs` but not yet wired into a fast-refresh code path

### References

- [Waveshare 3.52" wiki](https://www.waveshare.com/wiki/3.52inch_e-Paper_HAT)
- [Waveshare Python reference driver](https://github.com/waveshare/e-Paper)

Hardware validation ongoing — opening as draft pending display refresh test on target hardware.

PR #255, opened 2026-04-11
feat: add SSD1680 (Waveshare 2.13" V4) and UC8253 (Waveshare 3.52") drivers

### LINKED ISSUE #207: epd2in13_v4 support?
Hello,

I have a new e-paper HAT (which I believe is EPD 2.13" **V4**) -- which the documentation claim is compatible with the V3 (but has a fast refresh feature):

image
(Ref: https://www.waveshare.com/wiki/2.13inch_Touch_e-Paper_HAT_Manual#Version_Description)

However, when attempting to run the `epd2in13_v2` example (with the `epd2in13_v3` feature enabled) on the new board I'm running into several problems.

The first is PIN configuration.  I attempted to update the `epd2in13_v2` example to use the pins listed here:

swappy-20240717_102113

(Ref: https://www.waveshare.com/wiki/2.13inch_Touch_e-Paper_HAT_Manual#Hardware_Connection)

However this results in an "invalid argument" error just naively using the pin numbers.  Some [references online](https://stackoverflow.com/questions/40131771/sysfs-interface-i-cant-export-gpio-pins-in-a-xilinxs-board-zybo-and-other) suggest that there may be an offset for enabling/exporting GPIO pins?  Using this information I was able to find that my GPIO device has an offset of 512:

```
$ cat /sys/class/gpio/gpiochip512/base
512
```

This avoids the "invalid argument" errors and the example runs, but nothing is displayed on the epd2in13_v4 device:

```diff
iff --git a/examples/epd2in13_v2.rs b/examples/epd2in13_v2.rs
index b3cb64e..cff6c88 100644
--- a/examples/epd2in13_v2.rs
+++ b/examples/epd2in13_v2.rs
@@ -36,25 +36,25 @@ fn main() -> Result<(), SPIError> {
     spi.configure(&options).expect("spi configuration");

     // Configure Digital I/O Pin to be used as Chip Select for SPI
-    let cs = SysfsPin::new(26); //BCM7 CE0
+    let cs = SysfsPin::new(512 + 24);
     cs.export().expect("cs export");
     while !cs.is_exported() {}
     cs.set_direction(Direction::Out).expect("CS Direction");
     cs.set_value(1).expect("CS Value set to 1");

-    let busy = SysfsPin::new(24); // GPIO 24, board J-18
+    let busy = SysfsPin::new(512 + 18);
     busy.export().expect("busy export");
     while !busy.is_exported() {}
     busy.set_direction(Direction::In).expect("busy Direction");
     //busy.set_value(1).expect("busy Value set to 1");

-    let dc = SysfsPin::new(25); // GPIO 25, board J-22
+    let dc = SysfsPin::new(512 + 22);
     dc.export().expect("dc export");
     while !dc.is_exported() {}
     dc.set_direction(Direction::Out).expect("dc Direction");
     dc.set_value(1).expect("dc Value set to 1");

-    let rst = SysfsPin::new(17); // GPIO 17, board J-11
+    let rst = SysfsPin::new(512 + 11);
     rst.export().expect("rst export");
     while !rst.is_exported() {}
     rst.set_direction(Direction::Out).expect("rst Direction");
```

Do you have any advice on how I can debug things further.  Does this offset thing even make sense?  I'm happy to do the work to update the library to work with the device, but some pointers on where to start would be very helpful.  Thank you!

cogwheel886:
## Summary

Adds two new e-paper display drivers:

- **Waveshare 2.13" EPD V4** — SSD1680 controller, 250x122px, three-state partial refresh (full to base to partial), optional PWR pin (GPIO18)
- **Waveshare 3.52" EPD** — UC8253 controller, 360x240px, full and DU (quick) refresh via RefreshLut, alternating lut_flag sequence

Consolidates and supersedes #254 which had the 3.52" driver in an earlier state without hardware-verified fixes.

## Hardware tested

- Waveshare 2.13" e-Paper HAT V4 (SSD1680) on Raspberry Pi Zero 2WH (aarch64)
- Waveshare 3.52" e-Paper HAT (UC8253) on Raspberry Pi 5 (aarch64)

Both running Raspberry Pi OS (Bookworm, 64-bit).

## Examples

- epd2in13_v4.rs -- Ferris walking demo demonstrating SSD1680 partial refresh: Ferris the Rustacean walks left/right across the panel with no full-refresh flash between steps
- epd2in13_v4_status.rs -- system status display for the 2.13" panel (hostname, IP, CPU temp, RAM, disk, uptime, optional PiSugar 3 battery). Demonstrates the three-state partial refresh sequence in a real deployment. PiSugar battery status degrades gracefully to N/A without the daemon.
- epd3in52.rs -- carcinisation demo demonstrating UC8253 DU (quick) refresh: 8-frame deterministic cross-dissolve from a lobster to Ferris the Rustacean, with "Carcinisation" and "Rustacean" captions
- epd3in52_status.rs -- system status display for the 3.52" panel (hostname, IP, CPU temp, RAM, disk, uptime, optional PiSugar 3 battery)

All examples use gpio_cdev rather than the deprecated sysfs_gpio, required for Raspberry Pi OS Bookworm and Pi 5 compatibility.

## Implementation notes

- src/interface.rs gains soft_reset() -- RST LOW to HIGH with no trailing 200ms delay, used by SSD1680 partial refresh path
- epd3in52::update_partial_frame delegates to update_frame -- UC8253 has no hardware windowing; DU mode via RefreshLut::Quick is the correct fast-refresh path
- UC8253 reset timing corrected to 200ms/10ms matching the Waveshare Python reference driver (was 30us/10us)
- SSD1680 DisplayUpdateControl2 bytes annotated (0xF7/0xC7/0xFF)
- UC8253 init register sequence annotated against Python reference driver

## Checklist

- [x] Tested on real hardware (both displays)
- [x] cargo fmt applied
- [x] cargo clippy clean (3 pre-existing upstream warnings in src/graphics.rs and src/color.rs -- empty diff vs upstream/main)
- [x] cargo test -- 58 passed, 0 failed
- [x] Feature flags added (epd2in13_v4, epd3in52)
- [x] Examples registered in Cargo.toml with required-features
- [x] Closes #207


przemykomo:
Are you able to test it on a different hardware than raspberry pi? I can't seem to replicate the example with esp32c3 and the 3.52" screen.

cogwheel886:
> Are you able to test it on a different hardware than raspberry pi? I can't seem to replicate the example with esp32c3 and the 3.52" screen.

The driver itself (src/epd3in52/) is no_std compatible and should 
work on ESP32C3 with an embedded-hal 1.0 SPI and GPIO implementation. 
The examples (epd3in52.rs, epd3in52_status.rs) are Linux-specific — 
they use gpio_cdev, /dev/spidev0.0, and /proc for system stats, so 
they won't compile for a bare metal target.

For ESP32C3 you would need to wire up the driver directly using 
esp-hal's SPI and GPIO types. The driver's generic bounds are:
  SPI: SpiDevice
  BUSY: InputPin
  DC: OutputPin  
  RST: OutputPin
  DELAY: DelayNs

These are all embedded-hal 1.0 traits that esp-hal implements. The 
init sequence and display_frame() calls should work the same way — 
the UC8253 doesn't know what host is driving it.

dkess:
Hi, thanks for contributing this. I’ve tested the UC8253 driver on non-Raspberry-Pi hardware: a Seeed Studio XIAO RP2350 with the Waveshare 3.52" HAT/SKU 22698, using bare-metal Rust, `embassy-rp 0.10`, and the driver’s `embedded-hal 1.0` interfaces. The panel is connected in 4-wire SPI mode (`BS=0`), SPI mode 0 at 4 MHz, with BUSY active-low. I tested PR commit `a301138`.

The following comments are AI-generated, however I promise I am a real human and I did get this working on my own real hardware with some AI-discovered fixes:

> The generic driver works on RP2350, including GC and DU refreshes and sleep/wake, but I found two UC8253 details that were necessary on this hardware:
>
> **1. Reset LOW duration**
>
> With the current reset call:
>
> ```rust
> self.interface.reset(delay, 200_000, 10_000);
> ```
>
> BUSY remains low after initialization and the panel never refreshes. Changing only the LOW pulse to 2 ms fixes it:
>
> ```rust
> self.interface.reset(delay, 200_000, 2_000);
> ```
>
> I also reproduced this with a raw Rust port of Waveshare’s initialization/refresh sequence: 2 ms works, while 10 ms produces the same stuck-BUSY behavior as the driver. Waveshare’s official C driver uses `200 ms HIGH → 2 ms LOW → 200 ms HIGH`, and its Python driver also uses a 2 ms LOW pulse. The official C driver works on the same panel and wiring.
>
> I realize 10 ms worked on the Pi 5 used to test this PR, so this may be a hardware/platform-sensitive behavior rather than a universal failure, but using the vendor’s 2 ms timing also makes the driver work on RP2350.
>
> **2. Register `0x50` after the initial refresh**
>
> `init()` sets VCOM/data interval register `0x50` to `0xB7`. Waveshare’s demo changes it to `0x17` after the initial clear/refresh, but the current driver never makes that change. On my panel the first refresh succeeds, but subsequent refreshes show incorrect colors; in a black → white test, the panel ended black.
>
> I fixed that locally by tracking whether a refresh has completed since `init()`, writing:
>
> ```rust
> self.interface
>     .cmd_with_data(spi, Command::VcomDataSetting, &[0x17])?;
> ```
>
> after the first completed refresh, and resetting the state in `init()` so sleep/wake behaves like a fresh initialization.
>
> With those two changes I’ve run multiple rounds of:
>
> - 6 full/GC refreshes
> - 6 quick/DU refreshes
> - GC cleanup
> - sleep → wake → GC
>
> with alternating black/white half-screen patterns, and all refreshes have the expected colors. GC takes about 1.25 s and DU about 0.65 s including frame transfer and the existing 200 ms settle delay.
>
> One related observation: because `display_frame()` waits indefinitely for BUSY, the reset failure hard-blocked Embassy’s single-threaded executor. I used a BUSY timeout while diagnosing it, although that seems like a separate API/design question from these two small UC8253 changes.
>
> Thanks for adding this driver — with these changes it is working as a bare-metal `embedded-hal` driver on RP2350.

